PR #7, opened 2021-03-10
docs: update installation and quick start

kevinhermawan:
This PR updates installation and quick start on documentation.


docs/src/pages/quick-start.md
```diff
@@ -3,72 +3,78 @@ id: quick-start
...
 }
-
-render(<App />, document.getElementById('root'))
 ```
 
-These three concepts make up most of the core functionality of HTTPProbe. The next sections of the documentation will go over each of these core concepts in great detail.
+When you have created the configuration file, you can run monika as follows
+
+```
+monika -c <path_to_your_configuration.json>
```
dennypradipta, line 79: I think we should document that if there is no `config` flag (`-c`), it will automatically read './config.json'

dennypradipta (changes requested):

docs/src/pages/quick-start.md
kevinhermawan: Nice suggest!

gsukanto (approved):

sapiderman (approved):

dennypradipta (approved):
That's more like it! :+1:

PR #8, opened 2021-03-10
Doc/add probe doc

sapiderman:
# PR: Add Documentation

1. Add probee documentation
2. Update Readme to link with additional docs
3. Change example so that  probe can run on actual url


gsukanto (approved):

docs/src/pages/guides/probes.md
```diff
@@ -1,31 +1,73 @@
...
+      "name": "Name of the probe 2",
+      "description": "Probe to check GET health",
+      "request": { },
+      "alerts": []
+    }
+  ]
+```
 
-> To change this behavior, you can configure your queries both globally and per-query using the `staleTime` option. Specifying a longer `staleTime` means queries will not refetch their data as often
+Monica goes through each probe object, sends it out, and determines whether an alert or notification need to be sent out.
```
kevinhermawan, line 24: `Monika`

newm4n (approved):
LGTM

PR #37, opened 2021-03-15
Chore: Change example, test, and documentation domain to example.com

haricnugraha:
example.com domain is intended for documentation and also described in [RFC 2606](https://tools.ietf.org/html/rfc2606) and [RFC 6761](https://tools.ietf.org/html/rfc6761). It makes user understand that the domain is only example, and it prevents accidentally get email or traffic from Monika.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -5,7 +5,7 @@
       "id": "unique-id-mailgun",
       "type": "mailgun",
       "data": {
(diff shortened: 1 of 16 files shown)


raosan (approved):
lgtm

sapiderman (approved):
good catch!

Issue #22, opened 2021-03-12
Up and down threshold


newm4n:
Now, Every time the probe detected a "fail" in each of its detected,  Monika will send email notification, this could lead to thousand of annoying email. Thus,

When the a probe is in "SERVER_UP" status, if the probe detect 'n' times consecutive "fails" it will change status to "SERVER_DOWN". Then 'n' should be configurable per probe. Each time the probe enter "SERVER_DOWN" status, only 1 email should be sent, and no email should be sent afterward as long as the status not changed.

When the a probe is in "SERVER_DOWN" status, if the probe detect 'm' times consecutive "success" it will change status to "SERVER_UP". Then 'm' should be configurable per probe. Each time the probe enter "SERVER_UP" status, only 1 email should be sent, and no email should be sent afterward as long as the status not changed.

newm4n:
Example 

```
Assuming DOWN threshold = 3
Assuming UP threshold = 3

1 : U <--- UP  ----> Notif
2 : D <--- UP
3 : U <--- UP
4 : D <--- UP
5 : D <--- UP
6 : D <--- DOWN ----> Notif
7 : D <--- DOWN
8 : D <--- DOWN
9 : U <--- DOWN
10 : U <--- DOWN
11 : U <--- UP  ----> Notif

```

PR #40, opened 2021-03-16
Add Threshold to Limit How Many Failures/Success Needed to Send Notifications

dennypradipta:
# Monika PR
Quoted from #22 :
"Now, Every time the probe detected a "fail" in each of its detected, Monika will send email notification, this could lead to thousand of annoying email."

It will detect how many consecutive failures/success when probing a URL, then it will check if the threshold value is the same as consecutive failure/success times, it will send notification. 

Basically, the algorithm works like this:
- If current alert (e.g Status Not 2xx) is currently not true (meaning it's 2xx), x times failing request will set it to true and send notifications about a new alert has been detected.
- If current alert (e.g Status Not 2xx) is currently true (meaning it's 301, 404, 502, etc.), x times successful request will set it to false and send notification about your alert has been resolved.

## How to test
1. Use https://github.com/hyperjumptech/httptarget to simulate response time
2. Use this configuration:
```
{
  "interval": 5,
  "notifications": [
     <insert_your_credentials>
  ],
  "probes": [
    {
      "id": "1",
      "name": "Example",
      "description": "Probe",
      "request": {
        "method": "GET",
        "url": "http://localhost:51423/delay/100/300",
        "timeout": 7000,
        "headers": {
          "Authorization": ""
        },
        "body": {
          "username": "someusername",
          "password": "somepassword"
        }
      },
      "threshold": 2,
      "alerts": ["status-not-2xx", "response-time-greater-than-200-ms"]
    }
  ]
}

```
3. Run the binary

## Additional Context

Here are the logs when running MONIKA using the configuration above: https://gist.github.com/dennypradipta/635f25358afe4a160c4d94e98989ddd4

Screenshots of sent emails:
image
image


src/utils/get-server-status.ts
```diff
@@ -0,0 +1,180 @@
...
+}: {
+  errorName: string
+  probeStatus: StatusCount
+  validation: ValidateResponseStatus
+  config: Config
+  request: Probe['request']
+  threshold: number
+}) => {
+  log(`Checking for ${errorName} in probe...`)
+  if (probeStatus.isDown === false) {
```
gsukanto, line 28: Kindly reduce nested ifs in here.

src/utils/get-server-status.ts
```diff
@@ -0,0 +1,180 @@
...
+      // Add success count
+      probeStatus.consecutiveFailures = 0
+      probeStatus.totalSuccesses += 1
+      probeStatus.consecutiveSuccesses += 1
+
+      // If count is larger or equal than threshold...
+      // Set status to down, and reset consecutive failure
+      if (probeStatus.consecutiveSuccesses === threshold) {
+        log(`Sending resolved notification for "${errorName}"\n`)
+        await sendAlerts({
```
gsukanto, line 71: The function name is "calculate", why do we have side effect such as sendAlerts in here?

src/utils/looper.ts
```diff
@@ -13,12 +14,15 @@ async function doProbes(config: Config) {
...
-      log('id:', item.id, '- status:', probRes.status, 'for:', item.request.url)
+      log(
+        `id: ${item.id} - status: ${probRes.status} for: ${item.request.url} -- ${probRes.config.extraData?.responseTime}`
+      )
 
-      await sendAlerts({
-        validations: validatedResp,
-        notifications: config.notifications,
-        url: item.request.url ?? '',
+      await getServerStatus({
```
gsukanto, line 21: I believe this changes because we will send alert by using `calculateSuccessFailure` function called by `getServerStatus`. Can we create `getServerStatus` as dumb as get status only? So there is no side effect like sendAlerts?

src/utils/alert.ts
```diff
@@ -109,84 +127,78 @@ export const sendAlerts = async ({
...
+          } as WebhookData).then(() => ({
+            notification: 'slack',
+            alert: validation.alert,
+            url,
+          }))
+        }
+        case 'smtp': {
+          const transporter = createSmtpTransport(notification.data as SMTPData)
+          return sendSmtpMail(transporter, {
+            from: 'http-probe@hyperjump.tech',
```
gsukanto, line 187: Can we put TODO comment or read this configuration from env variable?

src/interfaces/probe.ts
```diff
@@ -1,9 +1,10 @@
 import { RequestConfig } from './request'
 
 export interface Probe {
-  id: number
+  id: string
   name: string
   description?: string
   request: RequestConfig
+  threshold: number
```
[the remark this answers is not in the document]
dennypradipta: Changed it to triggerThreshold

src/interfaces/server-status.ts
```diff
@@ -0,0 +1,16 @@
+export interface ServerStatus {
+  id: string
+  name: string
+  requests: number
+
+  isStatusCodeError: StatusCount
```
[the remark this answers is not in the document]
dennypradipta: Changed it to statusCodeError and responseTimeError

src/interfaces/server-status.ts
```diff
@@ -0,0 +1,16 @@
...
+  isResponseTimeError: StatusCount
+}
+
+export interface StatusCount {
```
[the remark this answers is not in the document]
dennypradipta: Changed it to Status

src/utils/get-server-status.ts
dennypradipta: It's kinda hard for me to reduce nested ifs. I use switch case. Is it okay?

src/utils/get-server-status.ts
dennypradipta: Moved out the sendAlerts to looper

src/utils/looper.ts
dennypradipta: Done

src/utils/alert.ts
dennypradipta: Done

src/interfaces/probe.ts
```diff
@@ -1,9 +1,10 @@
 import { RequestConfig } from './request'
 
 export interface Probe {
-  id: number
+  id: string
   name: string
   description?: string
   request: RequestConfig
+  triggerThreshold: number
```
gsukanto, line 8: What kind trigger threshold is this?
Is it for error case or success case or what?

src/interfaces/server-status.ts
```diff
@@ -0,0 +1,17 @@
+export interface ServerStatus {
+  id: string
+  name: string
+  requests: number
+
+  statusCodeError: Status
+  responseTimeError: Status
```
gsukanto, line 7: why do we need to have separate `statusCode` and `responseTime`?
What do you want to achieve?

And this status is not for error only, isn't this?

src/utils/get-server-status.ts
```diff
@@ -0,0 +1,181 @@
...
+    const results: Array<Status> = []
+
+    // Initialize server status
+    // This checks if there are no item in SERVER_STATUSES
+    // that doesn't have item with ID === id, push new ServerStatus
+    if (!SERVER_STATUSES.find((item) => item.id === id)) {
+      SERVER_STATUSES.push({
+        id,
+        name,
+        requests: 0,
```
gsukanto, line 101: can we put this static json element as constant?
example: `INIT_SERVER_STATUS`
So we can do concatenation such as:
```
SERVER_STATUSES.push ({
  id,
  name,
  ...INIT_SERVER_STATUS
})
```

src/utils/get-server-status.ts
```diff
@@ -0,0 +1,181 @@
...
+  validation: ValidateResponseStatus
+  threshold: number
+}) => {
+  log(`Checking for ${errorName} in probe...`)
+  switch (probeStatus.isDown) {
+    case false:
+      // If current probe is UP and trigger equals true
+      if (validation.status) {
+        // Add failure count
+        probeStatus.shouldSendNotification = false
```
gsukanto, line 28: Can we wrap/encapsulate, in separate function, for every state (combination of `isDown`, `validation.status` and `consecutiveSuccesses === threshold`)?

src/utils/get-server-status.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // Check if there is any alert that is triggered
+    // If the alert is being triggered <threshold> times, send alert and
+    // change the server status respectively.
+    const probeIndex = SERVER_STATUSES.findIndex((probe) => probe.id === id)
+    const currentProbe = SERVER_STATUSES[probeIndex]
+
+    // Add request count
+    currentProbe.requests += 1
```
gsukanto, line 128: This is one of the side effect, mutating data by adding counter.
Can we rename the function instead?

src/interfaces/probe.ts
dennypradipta: Currently, the threshold is for both success/error case.

src/utils/get-server-status.ts
dennypradipta: Sure

src/utils/get-server-status.ts
dennypradipta: Sure

src/utils/get-server-status.ts
dennypradipta: Sure thing

src/interfaces/server-status.ts
dennypradipta: It's for separating status code alert with response time alert. And it's to detect is it currently down/not

gsukanto (approved):

docs/src/pages/guides/probes.md
```diff
@@ -45,21 +45,24 @@ An actual probe request may be something like below.
...
+| method                      | Http method such as GET, POST, PUT, DELETE                                                                                     |
+| url                         | This is the url endpoint to dispatch the request to                                                                            |
+| timeout                     | Request timeout, after which time, `Monika`will assume timeout                                                                 |
+| headers                     | Your http header                                                                                                               |
+| body                        | Any http body if your method requires it                                                                                       |
+| successThreshold (optional) | How many times should a probe trigger send notifications and change the server to 'UP' if current server is 'DOWN' (default 5) |
+| failedThreshold (optional)  | How many times should a probe trigger send notifications and change the server to 'DOWN' if current server is 'UP' (default 5) |
```
nicnocquee, line 65: Number of times an alert should return false before Monika sends notifications. For example, when failedThreshold is 3, Monika will only send notifications when the probed URL returns status 2xx 3 times in a row. After sending the notifications, Monika will not send notifications anymore until the alert status changes. Default value is 5.

nicnocquee (changes requested):

nicnocquee (approved):

PR #48, opened 2021-03-18
chore: Change the link to download config sample from github

raosan:
Change url of example config file when there is no config file

### screenshot:
Screen Shot 2021-03-18 at 09 16 03


sapiderman (approved):
looks good

gsukanto (approved):

Issue #23, opened 2021-03-12
Probe detection record


newm4n:
Right now, for each of probe http call, the status and information is not recorded.
Thus is not possible to trace in historical manner of events and not able to see  to se trends in graphs fashion.

We need to save probe detection information which saves minimum.

1. Probe Name / ID
2. Time stamp
3. HttpRequest status (success or fail)
4. Response code
5. Response time

nicnocquee:
So that Monika is more flexible, 

- [ ] Add a flag to specify the directory to put the SQLite database file.

PR #49, opened 2021-03-18
Feat/presistent logs

sapiderman:
# PR Review

## What does this PR fixes

1. This PR implementes #23 
2. Historical logs are now stored in an sqlite db

## How was it implemented

1. Installing sqlite3
2. Created a table called history
3. Probe results are inserted into the logs

## How to test

1. run `bin/run`  should work. You should get statuscode sensitive coloring
2. run `bin/run -l` should dump the logs
3. run `bin/run --flush` should delete the logs. Only a "Y" character should be read as a yes.


src/index.ts
```diff
@@ -12,33 +13,67 @@ import { looper } from './utils/looper'
...
     version: flags.version({ char: 'v' }),
     help: flags.help({ char: 'h' }),
-    // flag with a value (-n, --name=VALUE)
+
     config: flags.string({
       char: 'c',
       description:
-        "JSON configuration file path e.g './config.json' (default './config.json')",
-      required: true,
+        'JSON configuration filename. If none is supplied, will look for default in current path.',
```
dennypradipta, line 34: I think we should add information about the "default file" name. Probably add (Default: './config.json') or similar?

src/utils/logger.ts
```diff
@@ -0,0 +1,47 @@
...
+      statusColor = 'white'
+  }
+
+  log(
+    'id:',
+    probe.id,
+    '- status:',
+    chalk.keyword(statusColor)(probRes.status.toString()),
+    'for:',
+    probe.request.url,
```
dennypradipta, line 42: I think it would be great to include response time as well.

`ID - Status: <status_code> For: <URL> -- <response_time>`

dennypradipta (changes requested):

src/index.ts
dennypradipta: So that user knows what file would be read as a default file

src/utils/looper.ts
```diff
@@ -42,18 +48,17 @@ async function doProbes(config: Config) {
...
     } catch (error) {
-      log(
-        'id:',
-        item.id,
-        '- status: ERROR for:',
-        item.request.url,
-        ':',
-        error.message
-      )
+      if (error) {
```
gsukanto, line 51: How come there is no error inside this catch?

src/utils/looper.ts
```diff
@@ -1,22 +1,28 @@
...
 
-async function doProbes(config: Config) {
+/**
+ * doProbe sends out the http request
+ * @param {object} config contains all the configs
+ */
+async function doProbe(config: Config) {
   // probe each url
   log('\nProbing....')
+  let probRes: AxiosResponseWithExtraData
```
gsukanto, line 18: Why we're defining the variable in here?
What is the problem with previous `const` variable?

src/utils/logger.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export async function probeLog(
+  probe: Probe,
+  probRes: AxiosResponseWithExtraData,
+  err: string
+) {
+  let statusColor: string
+
+  // colorize the statuscode
+  switch (Math.trunc(probRes.status / 100)) {
```
gsukanto, line 22: Can we wrap this switch case inside function?

src/index.ts
sapiderman: image

yup agree.

src/utils/history.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @param {object} probe is the probe config
+ * @param {object} probeRes this is the response time of the probe
+ * @param {string} errorResp if there was an error, it will be stored here
+ */
+export async function saveLog(
+  probe: Probe,
+  probeRes: AxiosResponseWithExtraData,
+  errorResp: string
+) {
+  const insertSQL = `INSERT into history (probe_id, status_code, probe_name, probe_url, response_time, error_resp) VALUES(?, ?, ?, ?, ?, ?);`
```
gsukanto, line 119: How about we add `created_at` column?

src/utils/history.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  let statusColor: string
+
+  await db.all(readRowsSQL, function (err: Error, data: any) {
+    if (err) {
+      log('warning: cannot read logfile. error:', err.message)
+    }
+    data.forEach((data: any) => {
+      // colorize the statuscode
+      switch (Math.trunc(data.status_code / 100)) {
```
gsukanto, line 68: Can we wrap this switch case inside function?

src/index.ts
```diff
@@ -12,33 +13,67 @@ import { looper } from './utils/looper'
...
+      printAllLogs()
+      closeLog()
+      return
+    }
+
+    if (flags.flush) {
+      const ans = await cli.prompt(
+        'Are you sure you want to flush all logs in history.db (Y/n)?'
+      )
+      if (ans === 'Y') {
```
gsukanto, line 62: Do we need to change `ans` into upper case?
in case user answering with `y` (lower case)?

src/index.ts
```diff
@@ -12,33 +13,67 @@ import { looper } from './utils/looper'
...
-        "JSON configuration file path e.g './config.json' (default './config.json')",
-      required: true,
+        'JSON configuration filename. If none is supplied, will look for default in current path.',
       default: './config.json',
     }),
-    // flag with no value (-f, --force)
-    force: flags.boolean({ char: 'f' }),
+
+    logs: flags.boolean({
+      char: 'l',
```
gsukanto, line 39: Can we put some comment why the char is `l`?

src/utils/logger.ts
sapiderman: oops good catch thanks. forgot it when refactored!

src/index.ts
sapiderman: Since this is a confirmation for a delete/flush ... a caps Y will ensure 2 two button press. so a definite YES want to delete logs. and prevents accidental deletion.

src/utils/history.ts
sapiderman: oooh had the table and column ready, not updated yet. fixed!

src/utils/looper.ts
sapiderman: fixed.

src/utils/looper.ts
sapiderman: ok. thanks!

src/utils/logger.ts
sapiderman: done.

src/utils/history.ts
sapiderman: done.. also moved from this module.. should never have been here.

src/index.ts
sapiderman: done.

src/utils/history.ts
```diff
@@ -0,0 +1,116 @@
...
+  responseTime: number
+}
+
+let db: any
+
+/**
+ * createTable will create the history table if it does not exist
+ */
+async function createTable() {
+  const createTablSQL = `CREATE TABLE IF NOT EXISTS history (
```
gsukanto, line 22: typo: `createTablSQL`, supposed to be `createTabelSQL`

gsukanto (approved):

src/utils/history.ts
sapiderman: fixed. thank you!

dennypradipta (approved):

PR #60, opened 2021-03-19
chore: add preinstall script to enforce npm version lower than 7

armedi:
### What issue is this PR trying to resolve?

In current development, the `package-lock.json` lockfile version keeps alternating between version 1 and 2.
This is because some developers are using the latest nodejs version (version 15) which comes with npm version 7, while most are using nodejs LTS (version 14) with npm version 6.

npm version 7 is not widely used yet, because most developers stick to LTS version for stability. In some cases, npm 6 users fail to install dependencies when lockfile version is 2

### How this PR addresses that issue?

- added `preinstall` script to check npm version, and abort the process if the version is greater than 6


nicnocquee (approved):

gsukanto (approved):
lgtm

Issue #38, opened 2021-03-16
WhatsApp Notification Channel


newm4n:
Now, Monika can not send notification through whatsapp.
In addition to email and slack channel, we should be able to send notification through whatsapp.
This configuration should be mentioned somewhere in the configuration json

PR #77, opened 2021-03-23
feature: whatsapp notification channel

ilmiawan:
# Monika PR
## What issue is this PR is trying to solve
Now, Monika can not send notification through whatsapp.
In addition to email and slack channel, we should be able to send notification through whatsapp.
This configuration should be mentioned somewhere in the configuration json

By adding a whatsapp notifications that can be set through config.json

## How to test
1. run unit test
2. set whatsapp config
3. run monika

related to issue #38 WhatsApp Notification Channel


src/utils/alert.ts
```diff
@@ -128,10 +134,10 @@ export const sendAlerts = async ({
   const ipAddress = getIp()
   const message = getMessageForAlert(validation.alert, url, ipAddress, status)
   const sent = await Promise.all<any>(
-    notifications.map((notification) => {
+    notifications.map(async (notification) => {
```
nicnocquee, line 137: No need to use async anymore here because the `notifications` is mapped to array of Promises and then the Promise.all will wait for them. 

So no need to add `await` on each cases below as well.

src/utils/alert.ts
```diff
@@ -145,49 +151,81 @@ export const sendAlerts = async ({
...
+
+          if (token) {
+            await Promise.all(
+              data.recipients.map(async (recipient) =>
+                sendTextMessage({
+                  recipient,
+                  token,
+                  baseUrl: data.url,
+                  message: validation.alert,
+                })
```
nicnocquee, line 219: Please refactor this. Create a function that can accept multiple recipients, or change so that `sendTextMessage` accepts array of recipients instead. Then we don't need to map the `data.recipients` here.

nicnocquee (changes requested):

config.example.json
```diff
@@ -41,6 +41,16 @@
       "data": {
         "url": "https://example.com/webhook"
       }
+    },
+    {
+      "id": "unique-id-whatsapp",
+      "type": "whatsapp",
+      "data": {
+        "recipients": ["08123456789"],
```
gsukanto, line 49: Phone number recipient is using local code (start with 0) instead of country code (ex: 62, 65, 1)?

src/utils/alert.ts
```diff
@@ -145,49 +151,81 @@ export const sendAlerts = async ({
...
+            url,
+          })
+        }
+        case 'whatsapp': {
+          const data = notification.data as WhatsappData
+          const token = await loginUser(data)
+
+          if (token) {
+            await Promise.all(
+              data.recipients.map(async (recipient) =>
```
gsukanto, line 213: why we're using await - promise.all - async in here?

src/utils/authorization.ts
```diff
@@ -0,0 +1,29 @@
+import { AxiosBasicCredentials } from 'axios'
+
+export const authBasic = (cred: AxiosBasicCredentials) => {
+  if (!cred.username)
+    throw new Error('Username should not be empty or undefined')
```
gsukanto, line 5: how about empty string?

src/utils/authorization.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!cred.password)
+    throw new Error('Password should not be empty or undefined')
```
gsukanto, line 7: how about empty string or password less than 6 char?

src/utils/alert.ts
ilmiawan: fixed

config.example.json
ilmiawan: fixed

src/utils/authorization.ts
ilmiawan: fix and updated

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

src/utils/authorization.ts
ilmiawan: this syntax will check undefined and empty string.

src/utils/authorization.ts
ilmiawan: fixed

sapiderman (approved):

gsukanto (approved):

src/utils/alert.ts
```diff
@@ -131,7 +137,7 @@ export const sendAlerts = async ({
     notifications.map((notification) => {
       switch (notification.type) {
         case 'mailgun': {
-          return sendMailgun(
+          sendMailgun(
```
nicnocquee, line 140: Why remove return? sendMailGun returns a Promise that will be caught by Promise.all.

nicnocquee (changes requested):

src/utils/alert.ts
ilmiawan: solved

nicnocquee (approved):

PR #78, opened 2021-03-23
ci: make sure the released version can be run before publishing

### LINKED ISSUE #73: Make sure the released version can be run before publishing
# Problem

The released version sometimes cannot be run due to various reasons, e.g., some files were not packed, etc.

# Solution

We need to make sure the released package can be run before publishing to npm. So we need to add a step to test the released version in the workflow npm-publish.yml. 

Testing steps:

- `npm pack`
- `npm install <path_to_tgz_file>`
- `monika -v`
- Check that the printed output is correct.

kevinhermawan:
This PR fixes #73

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -18,6 +18,17 @@ jobs:
       - run: npm ci
       - run: npm test
 
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm link
+      - run: monika -v
+
   publish-npm:
     needs: build
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -18,6 +18,17 @@ jobs:
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm link
+      - run: monika -v
```
nicnocquee, line 30: I might be wrong, but I don't think this will catch the errors/crashes of the released package. We need to actually pack the package and install then run the packed version as I mentioned in the issue #73 .

nicnocquee (changes requested):

nicnocquee (approved):

haricnugraha (approved):

Issue #63, opened 2021-03-19
Change notification in smtp to text and time to be known


ilmiawan:
**Describe the bug**
1. Currently the smtp notification set as html, therefore there is no formatting and become one line.
2. The time in notification is sent as unix format, not readable to human. 

**To Reproduce**
1. Set smtp config.json 
2. Run monika
3. Wait for notification firing
4. see the notification sent

**Expected behavior**
1. [x] the notifications sent as text.
2. [x] the time in notifications to be readable by human

nicnocquee:
I think we need to define the messaging for each of the alerts:

## status-not-2xx

On incident:

- title/subject: `[INCIDENT] Target URL is not OK`
- message: 
   ```
   Target $url is not healthy. It has not been returning status code 2xx $threshold times in a row.

   Alert: status-not-2xx
   Source: IP Address, $location_name
   Date: $date
   ```
   Example
    ```
   Target https://hyperjump.tech is not healthy. It has not been returning status code 2xx 5 times in a row.

   Alert: status-not-2xx
   Source: 192.168.1.1, Jakarta
   Date: Monday, 1 Jan 1970 18:00
   ```

On recovery:

- title/subject: `[RECOVERY] Target URL is not OK`
- message: 
   ```
   Target $url is back to healthy. 

   Alert: status-not-2xx
   Source: IP Address, $location_name
   Date: $date
   ```
   Example
    ```
   Target https://hyperjump.tech is back to healthy.

   Alert: status-not-2xx
   Source: 192.168.1.1, Jakarta
   Date: Monday, 1 Jan 1970 18:00
   ```

## response-time-greater-than-x-s

On incident:

- title/subject: `[INCIDENT] Target URL takes long to respond`
- message: 
   ```
   Target $url is not healthy. The response time has been greater than $time $threshold times in a row.

   Alert: response-time-greater-than-3-s
   Source: IP Address, $location_name
   Date: $date
   ```
   Example
    ```
   Target https://hyperjump.tech is not healthy. The response time has been greater than 3 seconds 5 times in a row.

   Alert: response-time-greater-than-3-s
   Source: 192.168.1.1, Jakarta
   Date: Monday, 1 Jan 1970 18:00
   ```

On recovery:

- title/subject: `[RECOVERY] Target URL takes long to respond`
- message: 
   ```
   Target $url is back to healthy.

   Alert: response-time-greater-than-3-s
   Source: IP Address, $location_name
   Date: $date
   ```
   Example
[...]

PR #81, opened 2021-03-24
Modify alert message to be more clear and send message as text

ilmiawan:
# Monika PR
There are some notification issue need to be solved :
1. Currently the smtp notification set as html, therefore there is no formatting and become one line.
2. The time in notification is sent as unix format, not readable to human.
3. need to define the messaging for each of the alerts

1. set smtp send message as text
2. change unix time format to be sent as readable time in UTC
3. define message as describe in issue #63 

## How to test
1. set smtp notifications in config
2. set probe test in config
3. run monika, and wait for the alerts to be sent to through smtp
4. the message should be shown as defined in issue #63 

e.g :
Subject : [INCIDENT] Target https://github.com takes long to respond
Body :
      Target https://github.com is not healthy. The response time has been greater than 400 ms 3 times in a row
      
      Time: Wed, 24 Mar 2021 05:32:06 GMT
      Target URL: https://github.com
      From server: 172.17.2.228


src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
+  ipAddress,
+  status,
+  trueThreshold,
+}: {
+  alert: string
+  url: string
+  ipAddress: string
   status: string
-): {
+  trueThreshold: number
```
gsukanto, line 86: can we change this variable name?
`trueThreshold` is too ambiguous.

src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
+
+      return `Target ${url} is not healthy.`
     }
-    return `New "${alert}" alert when probing "${url}".`
+
+    return `Target ${url} is back to healthy.`
   }
 
+  const now = Date.now()
+  const today = new Date(now)
```
gsukanto, line 124: Can we refactore line 123, 124, and 130 to be `const now = new Date().toUTCString()`?

image

src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
-): {
+  trueThreshold: number
+}): {
   subject: string
   body: string
 } => {
   const getSubject = (url: string, status: string) => {
-    if (status === 'UP') {
-      return `${alert} from probing ${url} has been resolved.`
+    if (alert === 'status-not-2xx') {
```
gsukanto, line 92: There are too many nested ifs. Kindly reduce the ifs.

src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
+      const statusAlert = `Target ${url} is not OK`
+      if (status === 'UP') {
+        return `[RECOVERY] ${statusAlert}`
+      }
+      return `[INCIDENT] ${statusAlert}`
+    }
+
+    const responseAlert = `Target ${url} takes long to respond`
+    if (alert === 'response-time-greater-than-400-ms' && status === 'UP') {
```
nicnocquee, line 101: What about `response-time-greater-than-1-s` or other time threshold?

src/utils/alert.ts
nicnocquee: if this is changed, don't forget to change in the other files. 

How about:

- `incidentThreshold` instead of trueThreshold
- `recoveryThreshold` instead of falseThreshold

nicnocquee (changes requested):

src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
 
   const getBody = (status: string) => {
-    if (status === 'UP') {
-      return `We found that your "${alert}" alert has been resolved when probing "${url}".`
+    if (status === 'DOWN') {
+      if (alert === 'status-not-2xx') {
+        return `Target ${url} is not healthy. It has not been returning status code 2xx ${trueThreshold} times in a row.`
+      }
+
+      if (alert === 'response-time-greater-than-400-ms') {
```
dennypradipta, line 113: Should use alert.includes() because the 400 ms is dynamic.

src/utils/alert.ts
```diff
@@ -72,34 +72,62 @@ export const validateResponse = (
...
+        return `Target ${url} is not healthy. The response time has been greater than 400 ms ${trueThreshold} times in a row`
```
dennypradipta, line 114: You can use `parseAlertStringTime()` to get the response time alert instead of defining 400ms

dennypradipta (changes requested):

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

src/utils/alert.ts
ilmiawan: fixed

nicnocquee (approved):

gsukanto (approved):

dennypradipta (approved):

PR #84, opened 2021-03-25
docs: add example to simulate HTML form submission POST request

raosan:
adding `Content-Type: application/x-www-form-urlencoded` in header example

Notes:
- Failed POST request with above header will still return 200 http code, so it is pure just to check wether the website is down or not. Since it will return 500 when the web is down.
- If we want to get the body right, then we need to change to stringify the body like this, but that's for another issue
```
var axios = require('axios');
var qs = require('qs');
var data = qs.stringify(config.body);
var config = {
  method: 'post',
  url: 'https://foo.com/login.php',
  headers: { 
    'Content-Type': 'application/x-www-form-urlencoded', 
  },
  data : data
};

```


docs/src/pages/guides/probes.md
```diff
@@ -67,8 +67,37 @@ Details of the field are give in the table below.
...
+        "headers": {
+          "Content-Type": "application/x-www-form-urlencoded"
+        },
+        "body": {
+          "username": "someusername",
+          "password": "somepassword"
+        }
+      },
+      "trueThreshold": 3,
+      "falseThreshold": 3,
```
raosan, line 96: to be in sync with PR #81 this need to be changed
```suggestion
      "incidentThreshold": 3,
      "recoveryThreshold": 3,
```

sapiderman (approved):

dennypradipta (approved):

nicnocquee (approved):

Issue #53, opened 2021-03-18
Config example needs to work out-of-the-box


sapiderman:
## Issue: Example config does not run without error

For best user experience, the config.json.example should work out of the box without error
Including:

1. notifications
2. probe target
3. alerts
4. intervals

update notifications documents also, according to the new config.json.example

nicnocquee:
I think the only way an example config works out of the box is if the config only contains webhook notification which points to a server we maintain. 

An idea to make Monika works _almost_ out of the box is to have multiple config examples where each of the example only contains single notification:

`config.smtp.gmail.example.json`: contains an `smtp` notification configuration as follows

```json
  "notifications": [
    {
      "id": "unique-id-smtp",
      "type": "smtp",
      "data": {
        "recipients": ["ENTER_YOUR_EMAIL_ADDRESS_HERE"],
        "hostname": "smtp.gmail.com",
        "port": 587,
        "username": "ENTER_YOUR_GMAIL_ADDRESS_HERE",
        "password": "ENTER_YOUR_GMAIL_PASSWORD_OR_APP_PASSWORD_IF_2FA_ACTIVE"
      }
    }
  ]
```

`config.sendgrid.example.json`: contains a `sendgrid` notification configuration as follows

```json
{
       "id": "unique-id",
       "type": "sendgrid",
       "data": {
         "recipients": ["ENTER_YOUR_EMAIL_ADDRESS_HERE"],
         "apiKey": "GET_API_KEY_FROM_SENDGRID"
       }
     }
```

And so on.

PR #98, opened 2021-03-28
docs: config example for each notification channel

raosan:
Implementing issue #53 by adding example file for each notification channel

Notes:
- In some feedback from user, it is mentioned that they think that they must use all notification channel. So we better remove the example that using all notification in the docs


haricnugraha (approved):

README.md
```diff
@@ -21,7 +21,14 @@ npm ci
 
 ## Getting Started
 
-To start monitoring URLs, you need to create a configuration file (JSON file). The configuration file contains the probes, alerts, and notification configurations. You can check the sample configuration file in [config.example.json](https://github.com/hyperjumptech/monika/blob/main/config.example.json) or from our [docs](https://github.com/hyperjumptech/monika/quick-start).
+To start monitoring URLs, you need to create a configuration file (JSON file). The configuration file contains the probes, alerts, and notification configurations. You can check the sample configuration file from our [docs](https://github.com/hyperjumptech/monika/quick-start) or copy from one of these files:
+
+- [config.smtp-gmail.example.json](https://github.com/hyperjumptech/monika/blob/main/config.smtp-gmail.example.json)
```
gsukanto, line 26: can we put those configs inside folder?
such as `config_sample`?

nicnocquee: agree

nicnocquee (approved):

Issue #123, opened 2021-04-01
Monika version used in docker is not the latest


armedi:
**Describe the bug**
The latest release of monika is version 1.3.0, but the version in dockerhub is 1.2.1

**To Reproduce**
1. `docker pull hyperjump/monika`
2. `docker run --rm -it hyperjump/monika sh`
3. `monika -v`
4. See the version of monika

**Expected behavior**
It is expected that the version is the latest (1.3.0 by the time of this report)

**Screenshots**
<img width="569" alt="Screen Shot 2021-04-01 at 16 20 26" src="https://user-images.githubusercontent.com/22452017/113273331-bd4da000-9306-11eb-994c-11e362dc1dc5.png">

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS]
 - Version [e.g. 11.1]

PR #130, opened 2021-04-06
fix: monika version in docker is not the latest

armedi:
This PR fixes issue #123 

The issue happened because docker publish and npm publish run around the same time, so the version of monika from npm got at the time of building image is still the previous one.

### How this PR solve that issue?

I changed the `npm install -g ...` process to install from source code, not from the npm repository

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node:14-alpine
 
 WORKDIR /monika
 
-COPY package*.json .
+COPY package*.json ./
 
 RUN npm ci
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ RUN npm ci
 
 COPY . .
 
+RUN npm run prepack
+
 RUN npm pack
 
 RUN npm install -g --unsafe-perm ./hyperjumptech-monika-*.tgz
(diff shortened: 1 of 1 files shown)


armedi:
rebased to main branch

Dockerfile
```diff
@@ -1,7 +1,19 @@
 FROM node:14-alpine
 
-RUN mkdir /config
+WORKDIR /monika
+
+COPY package*.json .
```
dennypradipta, line 5: This line gives me 
image

dennypradipta (changes requested):
image

Dockerfile
haricnugraha: I got same result.

armedi: fixed

sapiderman (approved):

dennypradipta:
Have you tried running your built docker image? I get this:
image

armedi:
> Have you tried running your built docker image? I get this:
> image

yes, I tried on macOS.
The `/lib` folder should be created by `npm pack` process. Let me check again

armedi:
> Have you tried running your built docker image? I get this:
> image

it is fixed

it looks like the `prepack` script does not get triggered during image building. so I added it manually

Screen Shot 2021-04-07 at 12 19 51

dennypradipta (approved):
All is working now :+1:

PR #152, opened 2021-04-19
Issue/136 example docs

### LINKED ISSUE #136: Create Examples section in the docs
To make it easier to user Monika, add an Examples section in the docs.

Screenshot 2021-04-08 at 10 16 53

Underneath this section, there should be pages about

- [ ] Monitoring a web page
- [ ] Monitoring a form submission
- [ ] Using chain request

dennypradipta:
This PR resolves #136 which is to create an "Examples" section in the docs

image

### DIFF
--- a/docs/src/pages/guides/examples.md
+++ b/docs/src/pages/guides/examples.md
@@ -1,8 +1,67 @@
 # Examples
 
+## Minimal Configuration
+
+Here is probe example with GET request to hit github.com:
+
+```json
+{
+  "probes": [
+    {
+      "requests": [
+        {
+          "url": "https://github.com"
+        }
+      ]
+    }
+  ]
+}
+```
+
+By default if you didn't define the method, it will be set as GET. Also, with this configuration you should be getting warnings for not defining notification channel and incident/recovery threshold.
+
+## Enabling Notification
+
+Here is probe example with GET request to hit Monika Landing Page:
+
+```json
+{
+  "notifications": [
+    {
+      "id": "unique-id-smtp",
+      "type": "smtp",
+      "data": {
+        "recipients": ["YOUR_EMAIL_ADDRESS_HERE"],
+        "hostname": "smtp.gmail.com",
+        "port": 587,
+        "username": "YOUR_GMAIL_ACCOUNT",
+        "password": "YOUR_GMAIL_PASSWORD_OR_APP_PASSWORD"
+      }
+    }
+  ],
+  "probes": [
+    {
+      "id": "1",
+      "name": "Monika Landing Page",
+      "description": "Landing page of awesome Monika",
+      "interval": 10,
+      "requests": [
+        {
+          "url": "https://hyperjumptech.github.io/monika",
+          "timeout": 7000
+        }
+      ],
+      "alerts": ["status-not-2xx"]
+    }
+  ]
+}
+```
+
+If the probe above has been alerted 5 times, which is the default incident/recovery threshold number, it will be sending a notification using the SMTP configuration. For more information about available notification channels, refer to [https://hyperjumptech.github.io/monika/guides/notifications](https://hyperjumptech.github.io/monika/guides/notifications).
+
 ## HTML-form-submission Example
 
-Here is probes example with POST request to simulate HTML form submission
+Here is probe example with POST request to simulate HTML form submission
 
 ```json
   "probes": [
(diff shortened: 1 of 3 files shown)


docs/src/pages/guides/examples.md
```diff
@@ -0,0 +1,246 @@
...
+      ],
+      "incidentThreshold": 3,
+      "recoveryThreshold": 3,
+      "alerts": ["status-not-2xx", "response-time-greater-than-2000-ms"]
+    }
+  ]
+}
+```
+
+In the configuration above, the first request will execute fetch all users available. If there are no triggered alerts, the response returned from the first request is ready to be used by the second request using values from `{{ responses.[0].data }}`. An example of the first request response should be like this:
```
nicnocquee, line 180: In the configuration above, the first request will fetch all users from `https://reqres.in/api/users`.  Then in the second request, Monika will fetch the details of the first user from the first request. If there are no triggered alerts, the response returned from the first request is ready to be used by the second request using values from `{{ responses.[0].data }}`.

Let's say the response from fetching all users in JSON format as follows:

docs/src/pages/guides/probes.md
```diff
@@ -146,110 +81,28 @@ Monika uses [Axios](https://github.com/axios/axios) to do requests, so the respo
...
-    ...
-  ]
-```
-
-So, in order to access the ID of the user, we need to define in the monika.json as `{{ responses.[0].data.data.[0].id }}` to get the first user ID from the first response. What if we want to get the `page` data? Simply just define it as `{{ responses.[0].data.page }}`.
-
-#### Pass Response Data as Headers value
-
-Here is an example of passing previous request(s) response into the headers:
+Details of the field are give in the table below.
```
nicnocquee, line 95: Details of the fields are shown in the table below.

docs/src/pages/guides/probes.md
```diff
@@ -146,110 +81,28 @@ Monika uses [Axios](https://github.com/axios/axios) to do requests, so the respo
...
-      "alerts": ["status-not-2xx", "response-time-greater-than-2000-ms"]
-    },
-  ]
-```
+| Topic      | Description                                                              |
+| :--------- | ------------------------------------------------------------------------ |
+| status     | HTTP Status Code (e.g 200, 403, 500)                                     |
+| statusText | HTTP Status Code Explanation (e.g OK, Forbidden, Internal Server Error)  |
+| config     | Request configuration (e.g URL, Method, Data, Headers, Body, etc.)       |
+| headers    | Response headers from the URL requested (e.g SetCookie, ETag, etc.)      |
```
nicnocquee, line 102: Response headers from the fetched request (e.g SetCookie, ETag, etc.).

nicnocquee (changes requested):

dennypradipta:
Thank you @nicnocquee for the thorough review. I have updated the PR :grinning:

nicnocquee (approved):

armedi (approved):

docs/src/pages/guides/examples.md
```diff
@@ -0,0 +1,249 @@
...
+        {
+          "url": "https://github.com"
+        }
+      ]
+    }
+  ]
+}
+```
+
+By default if you didn't define the method, it will be set as GET. Please note that with this configuration, you will not get any notifications when github.com is down since the notifications configuration is not defined.
```
sapiderman, line 21: since the notification configuration is not defined. (singular)

docs/src/pages/guides/examples.md
```diff
@@ -0,0 +1,249 @@
...
+      "recoveryThreshold": 3,
+      "alerts": ["status-not-2xx", "response-time-greater-than-200-ms"]
+    }
+  ]
+}
+```
+
+In the configuration above, Monika will first check `https://github.com/` then `https://github.com/hyperjumptech`. If the status code of `https://github.com/` is not 2xx (e.g., 200, 201), Monika **will not** check `https://github.com/hyperjumptech`.
+
+If there is a case where executing GET request to `https://github.com` triggered an alert, the next request will not be executed.
```
sapiderman, line 129: trigger**s** an alert. (soalnya pakai "is" di awal kalimat)

docs/src/pages/guides/examples.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Requests Chaining
+
+Monika supports requests chaining, which enables you to do multiple request and use previous request(s) response for other request. For example, after executing a GET request to certain API, the next request could use the previous request(s) response into their path/query parameters or headers.
```
sapiderman, line 133: Monika supports request chaining, which enables you to do multiple requests and the ability to use past responses from earlier requests. (s for tenses + compunds IS a pain in the ass)

sapiderman (approved):
looks good!

Issue #156, opened 2021-04-21
Allow config.json as default configuration file


nicnocquee:
To prevent breaking change, let's still allow `config.json` as default configuration file name.

- [x] When `-c` flag is not used, monika should search for monika.json in the current directory.
- [x] When there is no monika.json in the current directory, search for config.json.
- [x] When there is no config.json, show error.

PR #161, opened 2021-04-22
feat: Allow config.json as default configuration file

raosan:
implement issue #156 by adding `config.json` file checking in the directory


src/index.ts
```diff
@@ -40,6 +41,13 @@ import {
 import { notificationChecker } from './components/notification/checker'
 import { getConfig, setupConfigFromFile } from './components/config'
 
+function getDefaultConfig() {
+  const filesArray = fs.readdirSync('./')
+  const configJsonFile = filesArray.find((x) => x === 'config.json')
+
+  return configJsonFile ? `./${configJsonFile}` : './monika.json'
```
armedi, line 48: shouldn't monika.json take precedence over config.json?

raosan: With assumption that there are both file in the directory, the sequence is important. So maybe we should cater for that case too..

fixed btw.

nicnocquee (approved):

dennypradipta (approved):

Issue #147, opened 2021-04-16
Run a Probe only once, not by interval


newm4n:
Enable the user to make Monika to run a probe only once and terminate.
Instead of sending notification, Monika immediately give probe result as feedback to the user.

```
$ monika -c monika.json -once 1,3,5
Running Probe 1 - Testing User Dashboard
       Request 1 - POST https://abc.com/login [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
       Request 2 - GET https://abc.com/dashboard [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
Running Probe 3 - Testing Product Shopping Cart
       Request 1 - POST https://abc.com/login [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
       Request 2 - POST https://abc.com/addtocart [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
Running Probe 5 - Testing Requst Password
       Request 1 - POST https://abc.com/lostpasssword [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
       Request 2 - POST https://abc.com/confirmlost [OK]
                 Response Code : 200
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
       Request 3 - POST https://abc.com/changepassword [FAIL]
                 Response Code : 404
                 Response Time  : 420 ms
                 Response Size    : 253,274 bytes  
---
Running Probe 1 - Testing User Dashboard [OK]
Running Probe 3 - Testing Product Shopping Cart [OK]
Running Probe 5 - Testing Requst Password [FAIL]
$
```

sapiderman:
implementation is:
```
$ monika -c monika.json -r 1 -i 1,3,5  
$  monika --repeat 3 --ids 1,2,3,5,7
```

PR #182, opened 2021-05-09
feat: add repeats and custom id sequences

sapiderman:
### What this PR does
Implements #147 to add run options
1. Added -r  --repeat flag
2. Added -i  --id flag   
3. Add documentations in guides

### Implementation 
1. Added an id feeder to handle user specified probe.id sequences
2. Added counter checking for repeat control

## How to test
1. Running `monika` or `bin/run` should work normally
2. `monika -r x` where x is any number, should repeat just x times
3. `monika -i 1,2,3` should run probe id 1, 2 and then 3. Sequence can be any valid id sequence (ex: 1,2,3 or 3,2,1 or 1,2,1) 
4. `monika -r 3 -i 1,2` should run probe id 1 and 2 repeating 3 times
5. any invalid probe.id should be caught


src/index.ts
```diff
@@ -60,14 +60,21 @@ function getDefaultConfig() {
 class Monika extends Command {
   static description = 'Monika command line monitoring tool'
 
+  static examples = [
+    'monika',
+    'monika --1ogs',
```
armedi, line 65:
```suggestion
    'monika --logs',
```

src/index.ts
```diff
@@ -146,9 +165,7 @@ class Monika extends Command {
         const startupMessage = this.buildStartupMessage(config, flags.verbose)
 
         this.log(startupMessage)
-
-        // Loop through all probes
-        abortCurrentLooper = loopProbes(config)
+        idFeeder(config, Number(flags.repeat), flags.id)
```
armedi, line 168: when abortCurrentLooper is removed, monika will not be able to update it's configuration if new configuration is received from Symon

armedi (changes requested):

src/looper.ts
```diff
@@ -95,18 +95,17 @@ export function isIDValid(config: Config, ids: string): boolean {
   return true
 }
 
+let isAborted = false
```
armedi, line 98: moving this variable to global scope will cause the value of isAborted to stay `true` after it's first set to true. 

This means that when monika receive new configuration from Symon, it clear probe interval from the previous configuration and the new interval for the new configuration will also get aborted immediately. Or in other word, the next configurations will never run.

armedi (changes requested):

dennypradipta:
It seems like that some iteration did not run the probe according to the order:

1. Running with `./run -r 3`
image

In the first iteration it is being run using the correct order: 1,2,3. In the second iteration, it is being run in wrong order: 3,1,2

2. Running with `./run -3 -i 1,2`
image

In the first iteration it is being run using the wrong order: 2,1. In the second iteration, it is being run in correct order: 1,2

src/looper.ts
```diff
@@ -69,33 +70,120 @@ function sanitizeProbe(probe: Probe, index: number): Probe {
...
+  }
+  isAborted = false
 
   const abort = () => {
     isAborted = true
   }
 
+  // doing custom sequences?
+  if (ids) {
+    const idSplit = ids.split(',')
```
nicnocquee, line 163: need to trim the elements

nicnocquee (approved):

armedi (approved):

PR #217, opened 2021-06-10
Feat: reload probes on config change

armedi:
# Monika PR

Monika needs to reload when config file is changed

- Added file watcher using `chokidar`

## How to test
- start monika
- while monika is running, change the config file. You should see that monika reloads with new configuration


nicnocquee (approved):

sapiderman (approved):
this works! this is such a great idea why didn't we have it from the beginning!

PR #218, opened 2021-06-10
feat: read configuration from url

armedi:
# Monika PR

Monika should be able to fetch configuration from URL 

- update `config` flag to also accept URL as value
- fetch configuration from the supplied URL

## How to test
- start monika with this command
```sh
npm install
./bin/run -c https://raw.githubusercontent.com/hyperjumptech/monika/main/monika.example.json
```


src/components/config/index.ts
```diff
@@ -63,26 +64,31 @@ export const updateConfig = (data: Partial<Config>) => {
...
 }
+
+export const setupConfigFromFile = async (path: string) => {
+  const parsed = parseConfig(path)
+
+  await setupConfig(parsed)
+}
+
+export const setupConfigFromUrl = async (url: string) => {
+  const fetched = await fetchConfig(url)
```
nicnocquee, line 91: need to handle if the url is not found for example. right now, the error shown is just

```
 ›   Error: Request failed with status code 404
```

which is not so clear. The message should be

```
The configuration file in <insert_the_url_here> is unreachable. Please check the URL again or your internet connection. 
```

nicnocquee (approved):

src/components/config/index.ts
armedi: done

kevinhermawan (approved):

Issue #232, opened 2021-06-17
Check configuration from URL periodically


nicnocquee:
- [ ] Monika should check the config file in remote URL (if using remote URL config) periodically.
- [ ] The interval should be configured through the CLI flag `--config-interval` with values in seconds and default `900`.

PR #235, opened 2021-06-18
Added: check configuration from URL periodically

armedi:
This PR implements issue #232 

- [x] Monika should check the config file in remote URL (if using remote URL config) periodically.
- [x] The interval should be configured through the CLI flag --config-interval with values in seconds and default 900.

**How to test?**

- run `./bin/run --config <configuration url> --config-interval <chosen interval in seconds>`

### DIFF
--- a/src/components/config/index.ts
+++ b/src/components/config/index.ts
@@ -97,8 +97,17 @@ export const setupConfigFromFile = async (path: string, watch: boolean) => {
   }
 }
 
-export const setupConfigFromUrl = async (url: string) => {
+export const setupConfigFromUrl = async (
+  url: string,
+  checkingInterval: number
+) => {
   const fetched = await fetchConfig(url)
   await handshakeAndValidate(fetched)
   cfg = fetched
+
+  setInterval(async () => {
+    const fetched = await fetchConfig(url)
+    await handshakeAndValidate(fetched)
+    updateConfig(fetched)
+  }, checkingInterval * 1000)
 }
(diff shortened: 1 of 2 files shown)


nicnocquee (approved):

nicnocquee (commented):
~Monika doesn't seem to read the new probes when the file in remote URL is updated.~

nicnocquee:
There are some issues:

1. Seems like Monika will always think there is config update at least once even though the file is not changed.

To reproduce:

1. Create a valid monika.json
2. Serve it locally using serve (npm i -g serve): `serve .`
3. Run `./bin/run -c http://localhost:5000/secret.json --config-interval 30`

In the first round of the interval, it always thinks there is a new config.

Screenshot 2021-06-18 at 12 01 30

2. After Monika uses the new config, seems like some probing is not executed or something. It shows undefined in the log.

Screenshot 2021-06-18 at 12 02 44

nicnocquee (changes requested):
check the issue in the conversation

armedi:
> There are some issues:
> 
> 1. Seems like Monika will always think there is config update at least once even though the file is not changed.

Fixed
 
> 2. After Monika uses the new config, seems like some probing is not executed or something. It shows undefined in the log.

After doing some investigation, this also happens in the main branch. So I think this will need separate issue for further investigation and fix.

nicnocquee (approved):

kevinhermawan (approved):

Issue #233, opened 2021-06-17
Updated alerts is not displayed when config is updated with new alert options


armedi:
**Describe the bug**

When configuration is updated with new alerts option when monika is running, those new alerts option is not displayed correctly on log.

<img width="995" alt="Screen Shot 2021-06-17 at 15 48 46" src="https://user-images.githubusercontent.com/22452017/122364485-00bdbf00-cf84-11eb-8619-3b02ca086c3c.png">

In the image above, the alert is only displayed as "-"

Running the app in debugging mode, it looks that the updated alert option is not processed correctly into PROBE_STATUSES.

<img width="802" alt="Screen Shot 2021-06-16 at 16 11 16" src="https://user-images.githubusercontent.com/22452017/122365341-c86ab080-cf84-11eb-9da5-a5cc7268a285.png">

**To Reproduce**

1. run monika with this config
```json
{
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "url": "https://httpbin.org/status/200",
          "method": "GET"
        }
      ],
      "alerts": [
        "status-not-2xx",
        "response-time-greater-than-750-ms"
      ]
    }
  ]
}
```
2. While monika is running, update the alerts option:
```json
{
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "url": "https://httpbin.org/status/200",
          "method": "GET"
        }
      ],
      "alerts": [
        "status-not-2xx",
        "response-time-greater-than-1-ms"
      ]
    }
  ]
}
```
3. Look at the log, the "response-time-greater-than-1-ms" alert only displayed as "-"

**Expected behavior**

New alert options is displayed correctly

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS]
 - Version [e.g. 11.1]

PR #247, opened 2021-06-23
Fixed: Updated alerts is not displayed when config is updated with new alert options

armedi:
This PR fixes issue #233

## How does this PR fix the issue?

- reset saved probes status every time config changes

## How to test?

1. run monika with this config
```json
{
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "url": "https://httpbin.org/status/200",
          "method": "GET"
        }
      ],
      "alerts": [
        "status-not-2xx",
        "response-time-greater-than-750-ms"
      ]
    }
  ]
}
```
2. While monika is running, update the alerts option:
```json
{
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "url": "https://httpbin.org/status/200",
          "method": "GET"
        }
      ],
      "alerts": [
        "status-not-2xx",
        "response-time-greater-than-1-ms"
      ]
    }
  ]
}
```
3. Look at the log, the "response-time-greater-than-1-ms" alert should be displayed correctly

## Additional note:

In this PR I use rxjs to handle events with Observable. For now I mark it as draft because I need to hear feedback whether this is a right approach or not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3287,6 +3287,15 @@
           "integrity": "sha1-o7MKXE8ZkYMWeqq5O+764937ZU8=",
           "dev": true
         },
+        "rxjs": {
+          "version": "6.6.7",
+          "resolved": "https://registry.npmjs.org/rxjs/-/rxjs-6.6.7.tgz",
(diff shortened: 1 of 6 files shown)


armedi:
After a while I think

Issue #261, opened 2021-07-01
redundant message in CLI


sapiderman:
Unhelpful message "Response received"

version: on 1.4.0

## Screenshot
![image](https://user-images.githubusercontent.com/964106/124089700-56af5e00-da7e-11eb-85a2-377c6e3d3c9d.png)

sapiderman:
ok after quick investiagiton these are due to  some emmiter logs.  i think internal processes that do not impact user does not need to be printed. or maybe use log.debug() for it to come up only during dev.

PR #262, opened 2021-07-01
Cleanup log display in comand line (CLI)

sapiderman:
## PR
 Cleanup the log displays

## Implementation
1. Remove logs from event listener since it doesn't have actual functions yet.
2. Closing [261](https://github.com/hyperjumptech/monika/issues/261)

## How to test
1. run monika from source (bin/run)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -323,6 +323,21 @@ Please refer to the Monika documentations on how to how to configure notificatio
   }
 }
 
+// Subscribe FirstEvent
+em.addListener('TERMINATE_EVENT', function () {
+  // TODO: Add function here
+})
+
+// Subscribe to Sanitize Config
+em.addListener('SANITIZED_CONFIG', function () {
+  // TODO: Add function here
+})
+
+// Subscribe to Receive Response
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -323,21 +323,6 @@ Please refer to the Monika documentations on how to how to configure notificatio
   }
 }
 
-// Subscribe FirstEvent
```
nicnocquee, line 326: maybe just keep the listeners but remove the log.info so that in the future we just need to fill the functions.

nicnocquee (changes requested):

src/index.ts
```diff
@@ -323,21 +323,6 @@ Please refer to the Monika documentations on how to how to configure notificatio
...
-em.addListener('TERMINATE_EVENT', function (data) {
-  log.info('Monika Event: ' + data)
```
dennypradipta, line 328: Just add //TODO in case there is future events on developing this IMO

dennypradipta (changes requested):

dennypradipta (approved):

nicnocquee (approved):

Issue #242, opened 2021-06-23
validate response, send notification and store to db should be out of looper


ilmiawan:
As part of our plan to refactor Monika, we need to take out  validate response, send notification and store to db from looper.

<img width="849" alt="Screen Shot 2021-06-30 at 13 52 58" src="https://user-images.githubusercontent.com/8952493/123915342-986ed480-d9aa-11eb-811a-47fcbf4db7b8.png">

kevinhermawan:
This ticket is so big and mostly critical, I will separate this ticket into small PRs.

- [ ] Refactor validate response
- [ ] Refactor send notification
- [ ] Refactor store to db

PR #269, opened 2021-07-06
refactor: moves notification and store from probe to event emitter

kevinhermawan:
This PR fixes one of the issues in #242


src/index.ts
```diff
@@ -345,17 +355,125 @@ em.addListener('SANITIZED_CONFIG', function () {
...
+      em.emit(PROBE_SEND_NOTIFICATION, {
+        index,
+        probe,
+        status,
+        notifications,
+        totalRequests,
+        validatedResponseStatuses,
+      })
+
+      em.emit(PROBE_SAVE_LOG_TO_DATABASE, {
```
haricnugraha, line 416: You can just simply create sendNotification and saveLogToDatabase function, because I think there is no point to emit an event from here, at least for now.

src/index.ts
```diff
@@ -345,17 +355,125 @@ em.addListener('SANITIZED_CONFIG', function () {
...
+        index,
+        probe,
+        status,
+        notifications,
+      })
+    })
+})
+
+// PROBE_STATUS_PROCESSED -> PROBE_SEND_NOTIFICATION
+em.on(PROBE_SEND_NOTIFICATION, async (data: ProbeSendNotification) => {
```
dennypradipta, line 426: I think you should move it as a separate function, like `processProbeStatus()` or so, in case in the future there are going to be another function like:

```
em.on(PROBE_SEND_NOTIFICATION, async (data: ProbeSendNotification) => {
   processProbeStatus(data);
   // soonToBeDevelopedFunction(data);
   // anotherSoonToBeDevelopedFunction(data);
});
```

src/index.ts
kevinhermawan: Nice, thanks!

src/index.ts
```diff
@@ -345,17 +353,123 @@ em.addListener('SANITIZED_CONFIG', function () {
...
+      await probeSendNotification({
+        index,
+        probe,
+        status,
+        notifications,
+        totalRequests,
+        validatedResponseStatuses,
+      })
+
+      await probeSaveLogToDatabase({
```
haricnugraha, line 464: I don't think we need to await the `probeSendNotification` and `probeSaveLogToDatabase` function. Just catch the error and log it. It will make these functions executed asynchronously.

```javascript
      probeSendNotification({
        index,
        probe,
        status,
        notifications,
        totalRequests,
        validatedResponseStatuses,
      }).catch(err => log.error(err.message));

      probeSaveLogToDatabase({
        index,
        probe,
        status,
        notifications,
      }).catch(err => log.error(err.message));
```

kevinhermawan: Hmm, make sense. Thanks!

nicnocquee (approved):

dennypradipta (approved):

Issue #234, opened 2021-06-17
Integrate with prometheus


burnettk:
Heard about this project from stackshare.io. Have you considered exposing prometheus metrics from this thing? If you ran a little web server and exposed the relevant uptime/downtime data at a `/metrics` endpoint, anyone running monika in kubernetes could use prometheus and alertmanager to funnel alerts through their normal notification channels. Might be slick! Thanks!

PR #270, opened 2021-07-07
Added: Prometheus Metric Server

haricnugraha:
# Monika PR

## What issue does this PR address?

Add Prometheus metric server. Issue #234

## How did you implement it?

1. Set up a server.
2. Add `/metrics` endpoint that exposes Prometheus metrics.
3. Add `--prometheus` flag to enable the server.
4. Register the Prometheus metric collector.
5. Collect the metrics.
6. Expose the Monika probe request metrics along with [Prometheus default metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors) and [Node.js specific metrics.](https://github.com/siimon/prom-client/tree/master/lib/metrics)

### Available metrics

| Metric Name                          | Type      | Purpose                                      | Label                             |
| ------------------------------------ | --------- | -------------------------------------------- | --------------------------------- |
| `monika_probes_total`                  | Gauge     | collect total probe                          | -                                 |
| `monika_request_status_code_info`      | Gauge     | collect HTTP status code                     | id, name, url, method             |
| `monika_request_response_time_seconds` | Histogram | collect duration of probe request in seconds | id, name, url, method, statusCode |
| `monika_request_response_size_bytes`   | Gauge     | collect size of response size in bytes       | id, name, url, method, statusCode |

## How did you test?

1. Run Monika with the `--prometheus` flag with server port as a value. e.g., `./bin/run --prometheus 3001`.
2. Open `http://localhost:3001/metrics`.
3. [Download Prometheus](https://prometheus.io/download/).
4. Add text below to the [Prometheus configuration](https://prometheus.io/docs/prometheus/latest/getting_started/#configuring-prometheus-to-monitor-itself), under scrape_configs field

```yaml
- job_name: "monika"
  static_configs:
    - targets: ["localhost:3001"]
```

5. [Start Prometheus](https://prometheus.io/docs/prometheus/latest/getting_started/#starting-prometheus).
6. Run some [Prometheus query](https://prometheus.io/docs/prometheus/latest/querying/examples/). e.g., `monika_request_status_code_info`.

## Note

Here is the visualization of Monika Prometheus metrics from a [Grafana dashboard](https://grafana.com)
localhost_4000_d_3TUbRWinz_new-dashboard_orgId=1 refresh=1m


dennypradipta:
I had this error when running the ./run --prometheus 3001:
image

Here is my config:
image

src/plugins/metrics/prometheus/collector.ts
```diff
@@ -0,0 +1,107 @@
...
+  registerCollectorFromProbes(probes: Probe[]) {
+    // remove all registered metrics
+    register.clear()
+    this.requestMetricCollectors = []
+
+    probes.forEach((probe) => {
+      const { id, requests } = probe
+
+      requests?.forEach((_: RequestConfig, requestIndex: number) => {
+        const metricNamePrefix = `${snakeCase(id)}_request_${requestIndex}`
```
dennypradipta, line 50: This wouldn't work with IDs that uses UUID. Probably we could add handler if the ID is in UUID format, get the first part of the UUID?

Let's say the UUID is like "A-B-C-D" format. We could take the A only.

dennypradipta (changes requested):

src/plugins/metrics/prometheus/collector.ts
haricnugraha: Yeah, thank you for checking. I fixed it here [5da3ac9](https://github.com/hyperjumptech/monika/pull/270/commits/5da3ac91bd6e0c00df4ab4434fa3ca6ace2e71ae).

src/index.ts
```diff
@@ -118,6 +122,11 @@ class Monika extends Command {
       default: false,
     }),
 
+    prometheus: flags.integer({
+      char: 'p',
```
nicnocquee, line 126: - [ ] I think we can omit this. We might need the `-p` for something in the future. For now, user needs to use `--prometheus` to activate this feature.
- [ ] Also need to mention that this flag requires a number for the port. Probably need to mention the range of the port?
- [ ] Need to add `exclusive:` option as well to specify that this flag cannot be used with the `r` flag.

nicnocquee (changes requested):
How about unit/integration tests?

src/index.ts
haricnugraha: For the `p` flag, description about port, and exclusive flag, I fixed it here [d3a3b78](https://github.com/hyperjumptech/monika/pull/270/commits/d3a3b7819ed4711cb475d4e0357d912fbfdea209). I'm working on the test.

haricnugraha: I added the test here [883dd88](https://github.com/hyperjumptech/monika/pull/270/commits/883dd88fb8d9c7611f27a44f480a74e934163301).

nicnocquee (approved):

src/index.ts
```diff
@@ -126,6 +130,12 @@ class Monika extends Command {
       default: false,
     }),
 
+    prometheus: flags.integer({
```
dennypradipta, line 133: If the user does not specify the port number, at what port did prometheus start?

src/plugins/metrics/prometheus/publisher.ts
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      const prometheusMetrics = await register.metrics()
+
+      res.status(200).end(prometheusMetrics)
+    } catch (error) {
+      res.status(500).json({ message: error.message })
+    }
+  })
+
+  app.listen(port, () => {
```
dennypradipta, line 52: Because if I see this, there is no default port. Is this expected behavior?

src/index.ts
haricnugraha: If the user doesn't specify the port, Monika won't start. The user has to explicitly set the port. Do you have any suggestions?

sapiderman (approved):
looks good

src/index.ts
dennypradipta: I think it's better that way.

dennypradipta (approved):

bengoldenberg:
hey can you export the dashboard so others can import it? I didn't find any done dashboard around

bengoldenberg:
@sapiderman all the dashboard above will be amazing, but if i need to specify so the part of all probes

bengoldenberg:
> hi @bengoldenberg, what do you have in mind? you want to export the dashboard example above?
@sapiderman all the dashboard above will be amazing, but if i need to specify so the part of all probes

PR #291, opened 2021-07-23
Fixed(Config): write config file from har or postman

### LINKED ISSUE #282: Generate monika.json from a HAR file
Following up issue #274 , Monika should be able to generate a `monika.json` file from a HAR file by running

```
monika --create-config --har <path_to_har_file>
```

Note that Monika already has `--create-config` flag which will open our Config Generator in the browser. When this flag is combined with `--har` flag, it should just create a valid configuration file without running it.

ilmiawan:
# Monika PR
fix #282 Generate monika.json from a HAR file
Need to add capability to generate monika.json config file from har or postman
By writing to file the configuration parsed from har or postman if it has the flag `--create-config` while parsing from `--har` or `--postman` file flags.
## How to test
1. run `./bin/run --create-config --har ./test/testConfigs/harTest.har --output monika.config.json`
2. check on base folder `monika.config.json` file should be created


src/index.ts
```diff
@@ -153,6 +153,18 @@ class Monika extends Command {
...
+    output: flags.string({
+      char: 'o', // (o)utput file to write config to
+      description: 'Write monika config file to this file',
+      multiple: false,
+    }),
+
+    postman: flags.string({
+      char: 'p', // (p)ostman
+      description: 'Run Monika using a Postman json file.',
+      multiple: false,
```
nicnocquee, line 166: we need to add `exclusive` property to prevent this flag from being used with conflicting flags such as `--config`. Check about the exclusive property [here](https://oclif.io/docs/flags)

nicnocquee (changes requested):

src/index.ts
```diff
@@ -200,11 +212,8 @@ class Monika extends Command {
       startPrometheusMetricsServer(flags.prometheus)
     }
 
-    if (flags['create-config']) {
-      log.info(
-        'Opening Monika Configuration Generator in your default browser...'
-      )
-      await open('https://hyperjumptech.github.io/monika-config-generator/')
+    const isOpenConfigGenPage = await this.openConfigGenerator(flags)
```
gsukanto, line 215: If the variable name start with `is`, I would expect that variable will give me boolean value without side effect. Looks like the function `openConfigGenerator` has side effect (`open('https://hyperjumptech.github.io/monika-config-generator/')`).

src/index.ts
```diff
@@ -283,6 +292,19 @@ class Monika extends Command {
...
   }
 
+  async openConfigGenerator(flags: any): Promise<boolean> {
+    const openPage = flags['create-config'] && !flags.har && !flags.postman
+
+    if (openPage) {
+      log.info(
+        'Opening Monika Configuration Generator in your default browser...'
+      )
+      await open('https://hyperjumptech.github.io/monika-config-generator/')
```
gsukanto, line 302: cosmetic comment, feel free to ignore/resolve immediately: can we put `https://hyperjumptech.github.io/monika-config-generator/` inside constant variable?

src/components/config/index.ts
```diff
@@ -108,6 +109,18 @@ export const setupConfigFromFile = async (flags: any, watch: boolean) => {
       updateConfig(parsed)
     })
   }
+
+  if (flags['create-config'] && (flags.har || flags.postman)) {
+    const file = flags.output || 'monika.json'
```
gsukanto, line 114: Good idea to have default filename like `monika.json`. Cosmetic comment, feel free to ignore/resolve immediately: can default filename have better name to show that particular file is configuration for Monika?

src/index.ts
ilmiawan: done

src/index.ts
ilmiawan: since monika is to be run on client's engine this intended to be hardcoded.

src/components/config/index.ts
ilmiawan: this is due to the current system will search the name `config.json` or `monika.json` as default configuration file and changing this as default would need to change the configuration in config generator also.

src/index.ts
ilmiawan: fixed.

gsukanto (approved):

nicnocquee (approved):

PR #299, opened 2021-07-27
Functional Suitability : Add More Methods to Monika

### LINKED ISSUE #296: Monika method should not be limited to only POST and GET

dennypradipta:
# Monika Pull Request (PR)  

1. This PR fixed #296 

I refactored validate.ts to validate allowed probe request method. I copied the available methods from axios and map it into an array so that the probe request method will be validated with the methods array.

This PR also adds method name in the logs

## How to test  
1. Run monika with any configuration as usual
2. Run monika with one of the request is using OPTIONS method
3. Run monika with one of the request is using HEAD method

### DIFF
--- a/src/components/config/validate.ts
+++ b/src/components/config/validate.ts
@@ -43,7 +43,6 @@ import { Validation } from '../../interfaces/validation'
 import { isValidURL } from '../../utils/is-valid-url'
 
 const HTTPMethods = [
-  'CONNECT',
   'DELETE',
   'GET',
   'HEAD',
@@ -51,7 +50,9 @@ const HTTPMethods = [
   'PATCH',
   'POST',
   'PUT',
-  'TRACE',
+  'PURGE',
+  'LINK',
+  'UNLINK',
 ]
 
 const setInvalidResponse = (message: string): Validation => ({
@@ -265,7 +266,7 @@ export const validateConfig = (configuration: Config): Validation => {
         request.method = 'GET'
       }
 
-      if (HTTPMethods.indexOf(request.method) < 0)
+      if (HTTPMethods.indexOf(request.method.toUpperCase()) < 0)
         return PROBE_REQUEST_INVALID_METHOD
 
       // Check probe alert properties
--- a/src/components/config/validate.ts
+++ b/src/components/config/validate.ts
@@ -84,7 +84,7 @@ const PROBE_REQUEST_INVALID_URL = setInvalidResponse(
   'Probe request URL should start with http:// or https://'
 )
 const PROBE_REQUEST_INVALID_METHOD = setInvalidResponse(
-  'Probe request method is invalid!'
+  'Probe request method is invalid! Valid methods are GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS, PURGE, LINK, and UNLINK'
 )
 const PROBE_ALERT_INVALID = setInvalidResponse(
   `Probe alert should be 'status-not-2xx' or 'response-time-greater-than-<number>-(m)s`
--- a/src/components/config/validate.ts
+++ b/src/components/config/validate.ts
@@ -42,6 +42,18 @@ import { RequestConfig } from '../../interfaces/request'
 import { Validation } from '../../interfaces/validation'
 import { isValidURL } from '../../utils/is-valid-url'
 
+const HTTPMethods = [
+  'CONNECT',
+  'DELETE',
+  'GET',
+  'HEAD',
+  'OPTIONS',
+  'PATCH',
+  'POST',
+  'PUT',
+  'TRACE',
+]
+
 const setInvalidResponse = (message: string): Validation => ({
   valid: false,
   message: message,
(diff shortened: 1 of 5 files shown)


nicnocquee (approved):

src/components/probe/index.ts
```diff
@@ -66,6 +66,7 @@ export async function doProbe(
     id: probe.id,
     responseCode: 0,
     url: '',
+    method: 'GET',
```
sapiderman, line 69: Is the method purposedly prefilled with 'GET' and not read from probe.request[].method?

dennypradipta: Based on what I read, I think it was originally your code. 
I think what I want to achieve is that we should define the default mLog properties, in this case the method is 'GET'.

Because in the end, after all the probing has been done, it will be passed through the probeBuildLog:

armedi (approved):

src/components/config/validate.ts
```diff
@@ -42,6 +42,19 @@ import { RequestConfig } from '../../interfaces/request'
...
+  'DELETE',
+  'GET',
+  'HEAD',
+  'OPTIONS',
+  'PATCH',
+  'POST',
+  'PUT',
+  'PURGE',
+  'LINK',
+  'UNLINK',
```
gsukanto, line 55: Personally, I've never heard purge, link, and unlink method. Most of browser didn't use http method that is not officially in RFC. [This is http methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) used by mozilla (firefox).
Do we need to have those purge, link, and unlink methods?
Or can we just use [restful http methods](https://restfulapi.net/http-methods/) + options (preflight)?

src/components/config/validate.ts
```diff
@@ -71,7 +84,7 @@ const PROBE_REQUEST_INVALID_URL = setInvalidResponse(
   'Probe request URL should start with http:// or https://'
 )
 const PROBE_REQUEST_INVALID_METHOD = setInvalidResponse(
-  'Probe request method should be GET or POST only'
+  'Probe request method is invalid!'
```
gsukanto, line 87: Do we need more specific error message for this one? For example what is current method and what is available methods?

src/components/config/validate.ts
```diff
@@ -253,7 +266,7 @@ export const validateConfig = (configuration: Config): Validation => {
         request.method = 'GET'
       }
 
-      if (['GET', 'POST'].indexOf(request.method) < 0)
+      if (HTTPMethods.indexOf(request.method.toUpperCase()) < 0)
```
gsukanto, line 269: Smart!

src/interfaces/logs.ts
```diff
@@ -22,6 +22,8 @@
  * SOFTWARE.                                                                      *
  **********************************************************************************/
 
+import { Method } from 'axios'
```
gsukanto, line 25: image

Axios method doesn't have purge, link, and unlink.

src/interfaces/logs.ts
```diff
@@ -36,6 +38,7 @@ export interface LogObject {
   responseTime: number
   alert: Alerts
   notification: Alerts
+  method: Method | undefined
```
gsukanto, line 41: What is the case that can make `method` is undefined?

src/utils/pino.ts
```diff
@@ -54,7 +54,7 @@ const prettyPrint = {
       let alertMsg = ''
       let notifMsg = ''
 
-      const probeMsg = `${log.iteration} id:${log.id} ${log.responseCode} ${log.url} ${log.responseTime}ms`
+      const probeMsg = `${log.iteration} id:${log.id} ${log.responseCode} ${log.method} ${log.url} ${log.responseTime}ms`
```
gsukanto, line 57: cosmetic comment, feel free to ignore/resolve immediately: its kind of useless having `method` without `url`/endpoint in this log/message.

src/components/config/validate.ts
dennypradipta: I'm basically with you with the purge, link, unlink method. But when I import the Method type from axios, it is defined like this:

image

I just matched the array with the type

src/interfaces/logs.ts
dennypradipta: Related to other comment, the types are defined like this:
image

I posted the filename just in case you are wondering

src/interfaces/logs.ts
dennypradipta: Let's say that user defined the config using minimal configuration:

```
{
   "probes": [
     {
       "requests": [
         {
           "url": "https://github.com"
         }
       ]
     }
   ]
 }
```

They did not define the request method. So that's why when I didn't put `undefined` in the interface, this happens:

image

src/utils/pino.ts
dennypradipta: The URL is there? In ${log.url}?

src/components/config/validate.ts
dennypradipta: Should be fixed in the newest update! :D

gsukanto (approved):

src/utils/pino.ts
gsukanto: ups, my bad.

PR #302, opened 2021-07-27
Feature(config): [295] return after create configuration file

### LINKED ISSUE #295: Converting HAR should not run monika
**Describe the bug**
Monika is running the probes even though it should only create configuration file.

**To Reproduce**
1. Run `monika --create-config --har <path to a HAR>`
2. Monika is running even though it shouldn't. 

**Expected behavior**
When there is create-config, monika should just create config. no need to run.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 1.4.1

ilmiawan:
# Monika Pull Request (PR)  
fix #295 
Monika is running the probes even though it should only create configuration file.

fix logic for returning after create configuration file

## How to test  
1. run cmd `./bin/run --create-config --har copy_example.har`
2. monika should immediately return, and closed 
3. make sure monika is not running and monika.json file is created


src/components/config/index.ts
```diff
@@ -83,6 +85,18 @@ const handshakeAndValidate = async (config: Config) => {
   }
 }
 
+const writeConfigFile = (flags: any, parsedConfig: Config) => {
+  const file = flags.output || 'monika.json'
+  writeFile(file, JSON.stringify(parsedConfig), 'utf8', function (err) {
+    if (err) {
+      log.error('An error occured while writing har config to File.')
```
nicnocquee, line 92: Please change the message to 

```
Could not write the new configuration to ${file}.
```

Because the new config can also from postman, not only har.

src/components/config/index.ts
```diff
@@ -97,6 +111,11 @@ export const setupConfigFromFile = async (flags: any, watch: boolean) => {
   }
 
   const parsed = parseConfig(path, type)
+  if (flags['create-config'] && (flags.har || flags.postman)) {
+    writeConfigFile(flags, parsed)
+    return
+  }
+
```
nicnocquee, line 118: i think this shouldn't be here. it makes things complicated. check my comment in the setupConfig function.

src/components/config/index.ts
```diff
@@ -105,20 +124,13 @@ export const setupConfigFromFile = async (flags: any, watch: boolean) => {
...
-
-  if (flags['create-config'] && (flags.har || flags.postman)) {
-    const file = flags.output || 'monika.json'
-    writeFile(file, JSON.stringify(parsed), 'utf8', function (err) {
-      if (err) {
-        log.error('An error occured while writing har config to File.')
-        return log.error(err)
+      if (flags['create-config'] && (flags.har || flags.postman)) {
+        writeConfigFile(flags, parsed)
+        return
```
nicnocquee, line 129: i don't this is needed. this block is executed when there's a change in the file. but when `create-config` is specified, monika doesn't run or even watching the file change.

src/components/config/index.ts
```diff
@@ -138,3 +150,25 @@ export const setupConfigFromUrl = async (
...
 }
+
+export const setupConfig = async (flags: any): Promise<Config | undefined> => {
+  if (flags['create-config'] && !flags.har && !flags.postman) {
+    log.info(
+      'Opening Monika Configuration Generator in your default browser...'
+    )
+    open('https://hyperjumptech.github.io/monika-config-generator/')
+    return
+  }
```
nicnocquee, line 161: One thing that is for certain: **When `create-config` exists, monika does not run. It either opens the browser or generate a config file from postman or har.** So, it's simpler to put the logic here:

```javascript
// extract this as a function so can be called by other functions
const getPathAndTypeFromFlags = (flags: any): { path: string; type: string } => {
  let path = flags.config
  let type = 'monika'

  if (flags.postman) {
    path = flags.postman
    type = 'postman'
  }
  if (flags.har) {
    path = flags.har
    type = 'har'
  }

  return {
    path,
    type,
  }
}

export const createConfig = async (flags: any) => {
  if (!flags.har && !flags.postman) {
    log.info(
      'Opening Monika Configuration Generator in your default browser...'
    )
    open('https://hyperjumptech.github.io/monika-config-generator/')
  } else {
    const { path, type } = getPathAndTypeFromFlag(flags)

    const parsed = parseConfig(path, type)
    writeConfigFile(flags, parsed)
  }
}

export const setupConfig = async (flags: any) => {
  if (isUrl(flags.config)) {
    await setupConfigFromUrl(flags.config, flags['config-interval'])
  } else {
    const watchConfigFile = !(
      process.env.CI ||
      process.env.NODE_ENV === 'test' ||
      flags.repeat
    )

    await setupConfigFromFile(flags, watchConfigFile)
  }
}
```

With this, 
1. we can call `createConfig` from `src/index.ts` as early as possible, because this function does not require any db setup or anything else.
2. `setupConfig` function is responsible only for running monika using existing configuration file.

src/index.ts
```diff
@@ -214,26 +207,10 @@ class Monika extends Command {
...
-        const watchConfigFile = !(
-          process.env.CI ||
-          process.env.NODE_ENV === 'test' ||
-          flags.repeat
-        )
-
-        await setupConfigFromFile(flags, watchConfigFile)
+      await setupConfig(flags)
+      if (flags['create-config']) {
+        return
```
nicnocquee, line 213: we don't need to put this here. check my comment above. we can instead call `createConfig` as early as possible when the `run` function starts.

```javascript
  async run() {
    const { flags } = this.parse(Monika)

    if (flags['create-config']) {
      await createConfig()
      return
    }
```

src/components/config/index.ts
```diff
@@ -83,6 +85,18 @@ const handshakeAndValidate = async (config: Config) => {
   }
 }
 
+const writeConfigFile = (flags: any, parsedConfig: Config) => {
```
nicnocquee, line 88: can you change this function to async? so the caller can wait until the write succeeds.

nicnocquee (changes requested):

nicnocquee:
Please also add tests to make sure when `--create-config` and `--har` or `--postman` are specified, a new configuration file is created.

src/components/config/index.ts
ilmiawan: updated

src/components/config/index.ts
ilmiawan: removed

src/components/config/index.ts
ilmiawan: fixed refactor

src/components/config/index.ts
ilmiawan: removed

src/index.ts
ilmiawan: moved

src/components/config/index.ts
ilmiawan: updated

gsukanto (approved):

src/components/config/__tests__/create-config.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe('Har config', () => {
+  describe('Create config from har file', () => {
+    it('should create config from har file', async () => {
+      const flags = {
+        har: './src/components/config/__tests__/form_encoded.har',
+        output: 'monika.har.json',
+      }
+      await createConfig(flags)
+      expect(fs.lstatSync('monika.har.json').isFile()).to.be.true
```
nicnocquee, line 47: can we also add expectation to make sure that the generated monika.har.json is as expected? just compare the content of the generated json with the expected json.

src/components/config/__tests__/create-config.test.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('Postman config', () => {
+  describe('Create config from postman file', () => {
+    it('should create config from postman file', async () => {
+      const flags = {
+        postman:
+          './src/components/config/__tests__/simple.postman_collection.json',
+        output: 'monika.postman.json',
+      }
+      await createConfig(flags)
+      expect(fs.lstatSync('monika.postman.json').isFile()).to.be.true
```
nicnocquee, line 61: same as above.

nicnocquee (changes requested):

nicnocquee (approved):

PR #307, opened 2021-07-30
Fix: do not overwrite existing config

### LINKED ISSUE #306: Current configuration is forecebily overwritten
When using `--create-config` with `--har` the existing monika.json file that was lovingly crafted is overwritten without any warning.

Overwritting files without any warning or confirmation is not a good user experience.

Bring my file back!

ilmiawan:
# Monika Pull Request (PR)  
fix #306 
When using --create-config with --har the existing monika.json file that was lovingly crafted is overwritten without any warning.
Overwritting files without any warning or confirmation is not a good user experience.

1.  do not continue writing config file if it is already exists
2. overwrite config file if force flag is given

## How to test  
1. run monika using flags `./bin/run --create-config --har ./src/components/config/__tests__/form_encoded.har --output monika.config.json`, it should show error message if output file is already exist. e.g run 2 times using the same command
2. run monika using flags `./bin/run --create-config --har ./src/components/config/__tests__/form_encoded.har --output monika.config.json --force`, it should overwrite the existing config file. Should show message info `INFO: monika.config.json file has been created.`


src/components/config/index.ts
```diff
@@ -161,9 +161,22 @@ export const createConfig = async (flags: any) => {
...
+      log.error(`Couldn't found the ${path} file.`)
+      return
+    }
+
     const parsed = parseConfig(path, type)
     const file = flags.output || 'monika.json'
 
+    if (existsSync(file) && !flags.force) {
+      log.error(
+        `Cannot write configuration file to ${file} because file already exists.`
```
dennypradipta, line 175: I may need some feedback on this, but how about we use incremental filename?
https://stackoverflow.com/questions/41314177/how-to-increment-filename-if-file-exists

Let's say that monika.json is already exists, when there is no --force flag, it will check if there is any file named monika.json in the directory. 
- If it exists, create monika-1.json
- If it doesn't exists, proceed with monika.json
- If there is monika and monika-1.json, it will be monika-2.json
- If there is only monika-1.json, proceed with monika.json

But then again, it's OK if we stay with this

monika.config.json
```diff
@@ -0,0 +1,35 @@
...
+            "accept-language": "en-GB,en-US;q=0.9,en;q=0.8",
+            "cookie": "_fbp=fb.1.1624520390835.1717829411; _gid=GA825140508.1627380527; _gat_gtag_UA_89898_8=1; _ga_AAAAA=GS1.1.1.26.1.1.0; _ga=1.1"
+          },
+          "params": {},
+          "body": "email=myemail&password=mypass"
+        }
+      ]
+    }
+  ]
+}
```
armedi, line 35: I think this file is committed by mistake, is that true?

src/index.ts
```diff
@@ -171,6 +171,11 @@ class Monika extends Command {
       description: 'Write monika config file to this file',
       multiple: false,
     }),
+
```
sapiderman, line 174: i think adding a --force flag is not solving the user experience issue. 

A better experiece may be:
"file monika.json already exist, overwrite (n/Y)"  ==> defaults to no if any other key pressed
This continues the command and at the same time alerts the user.
If the user click No/cancle, print out a message that the "-o" output option exist.

Failing the command  "file already exist!" and then forching him (or me) to retype the whole long commadn again AND add a new flag is not a good experience. AND we have to add another flag.

sapiderman (changes requested):
i think is more elegant to ask the user overwrite: n/Y

src/components/config/index.ts
gsukanto: +1 with Denny idea

gsukanto (approved):

src/index.ts
gsukanto: Agreed.
This `force` command is too ambiguous. What is the thing that is forced?

src/components/config/index.ts
sapiderman: There's the -o option for this already. having number is ambiguous and hard to find.

src/index.ts
nicnocquee: You don't need to retype the whole long command, just press the up arrow keyboard 🤣  but I don't mind with the Y/n. But `force` is still needed when user runs the command in automated environment like CI.

dennypradipta: If force command is ambiguous, maybe we can change it to --force-overwrite

src/components/config/index.ts
```diff
@@ -161,9 +162,26 @@ export const createConfig = async (flags: any) => {
...
     const parsed = parseConfig(path, type)
     const file = flags.output || 'monika.json'
 
+    if (existsSync(file) && !flags.force) {
+      const ans = await cli.prompt(
+        `\n${file} file is already exists. Overwrite (Y/n)?`
+      )
+
+      if (ans.toLowerCase() !== 'y') {
+        log.warn(`Exit without overwriting ${file}.`)
```
sapiderman, line 180: Kang Mus maybe better for the exit message to be: 
Command cancelled. You can use the `-o` flag to specify an output file or `--force` to overwrite without prompting

sapiderman (approved):
small minor comment but looks good.

src/components/config/index.ts
ilmiawan: done

dennypradipta (approved):

PR #309, opened 2021-07-30
Added: Improve Alert/Notification messages sent to channels

### LINKED ISSUE #303: Improve Alert/Notification messages sent to channels
Improve the notification messages that is sent from Monika

It should contain: 
1. The type of error/notification 
2. And what specifically caused the notification

The following is the present notification:
image

It's a bit repetitive, and missing some info; what was the error code that triggered the not-2xx? 
Change time to "At". Time is the notification or the email? At is a definite when.

A better notifications might be:
  
Monika is running on LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname).

Monika is terminated on LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname).

**New Incident from Monika**

* Alert :  Status is **500**, was expecting **200**
* url : http://api.example.com/
* At : 2021-07-28 , 13:30:30
* Monika: LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname)
  

**New Incident from Monika**

* Alert : Response time is  **700ms** expecting a **200ms**
* url : http://api.example.com/
* At : 2021-07-28 , 13:30:30
* Monika: LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname)

**New Recovery from Monika**

* Alert : Service is ok. Status now **200** was **401**
* url : http://api.example.com/
* At : 2021-07-28 , 13:30:30
* Monika: LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname)
  

**New Recovery from Monika**

* Alert : Service is ok. Response now is within **50ms**
* url : http://api.example.com/
* At : 2021-07-28 , 13:30:30
* Monika: LOCAL_IP_ADDRESS (local), PUBLIC_IP_ADDRESS (public), HOSTNAME (hostname)

When public ip cannot be determined, don't include it, just use the local ip address and hostname.

raosan:
# Monika Pull Request (PR)  
Fix #303 

add more detailed notification:
- with current value and expected in a a string
- with ip local, ip public, and hostname

1.  refactor code,  add `responseValue` to `interface ValidateResponse`
2. get the string message
3. get public ip and hostname
4. fix bug: wrong notification to send in alert (in `src/index.ts` around line 525)
```
 statuses
     ?.filter((status) => status.shouldSendNotification)
     ?.forEach((status, index) => {
        probeSendNotification({
         probe,
         status,
         notifications,
         totalRequests,
         validatedResponseStatuses[index],
     }).catch((error: Error) => log.error(error.message))
```

## How to test  
1. run monika
2. wait until there is INCIDENT notification. 
3. it will add detailed information
4. currently implemented in Desktop and Teams notifications

Screen Shot 2021-08-02 at 15 23 51


src/utils/ip.ts
```diff
@@ -23,6 +23,10 @@
  **********************************************************************************/
 
 import { networkInterfaces } from 'os'
+import axios from 'axios'
+import { log } from './pino'
+
+export let publicIpAdress = ''
```
gsukanto, line 29: A little bit typo supposed to be: `publicIpAddress` or `publicIPAddress`

src/plugins/validate-response/checkers/index.ts
```diff
@@ -26,7 +26,9 @@ import { AxiosResponseWithExtraData } from '../../../interfaces/request'
 import responseTimeGreaterThanX from './res-time-greater-than-x'
 import statusNot2xx from './status-not-2xx'
 
+// parse string like "response-time-greater-than-200-ms" and return the time in ms
 export const parseAlertStringTime = (str: string): number => {
+  // match any string that ends with digits followed by unit 's' or 'ms'
```
gsukanto, line 31: Typo: `match any digit ...`

src/components/notification/alert-message.ts
```diff
@@ -66,6 +69,25 @@ export function getMessageForAlert({
     return `Target ${url} is back to healthy.`
   }
 
+  const getExpectedMessage = (status: string, responseValue: number) => {
+    if (alert === 'status-not-2xx' && status === 'DOWN') {
```
gsukanto, line 73: Minor suggestion: having nested ifs (in this case) is clearer/easier to read. For example:
```
if (alert === 'status-not-2xx') {
  if(status === 'DOWN') {
    return `Status is ${responseValue}, was expecting 200.`
  }

  if(status === 'UP') {
    return `Service is ok. Status now 200`
  }
}

if (alert.includes('response-time-greater-than-')){
  const alertTime = parseAlertStringTime(alert)

  if(status === 'DOWN') {
      return `Response time is ${responseValue}ms expecting a ${alertTime}ms`
  }

  if(status === 'UP') {
    return `Service is ok. Response now is within ${alertTime}ms`
  }
}
```

src/utils/ip.ts
raosan: nice catch

gsukanto (approved):

src/index.ts
```diff
@@ -175,6 +173,7 @@ class Monika extends Command {
 
   async run() {
     const { flags } = this.parse(Monika)
+    getPublicIp() // calling it here to give time getting public IP before calling on alert functions
```
nicnocquee, line 176: i think this should wait until resolve. also it should be after the createConfig, because createConfig does not need the ip address.

src/utils/ip.ts
```diff
@@ -40,3 +44,17 @@ export default function getIp(): string {
...
 }
+
+export function getPublicIp() {
+  axios({
+    method: 'GET',
+    url: 'https://ifconfig.co/ip',
+  })
+    .then((res) => {
+      log.info(`Monika is running on Public IP ${res?.data}`)
+      publicIpAddress = res?.data
```
nicnocquee, line 55: i think you should check first if res?.data has value before setting it to publicIpAddress and calling log.info

nicnocquee (changes requested):

src/index.ts
raosan: I have tried using async await, but the down side is the test will run longer because there are many test that running monika. What do you think?

src/utils/ip.ts
raosan: fixed

src/index.ts
dennypradipta: How about mock getIP when testing? Always return 192.168.0.1?

raosan: Nice suggestion. Will try it

src/utils/ip.ts
```diff
@@ -40,3 +44,19 @@ export default function getIp(): string {
...
+  })
+    .then((res) => {
+      if (res?.data) {
+        log.info(`Monika is running on Public IP ${res.data}`)
+        publicIpAddress = res.data
+      }
+    })
+    .catch(() => {
+      log.info(`Can't obtain Public IP`)
+    })
```
nicnocquee, line 61: I think we should use [stun](https://www.npmjs.com/package/stun) instead to get the IP address.

nicnocquee (changes requested):

src/index.ts
gsukanto: If you want to have mock IP, it would be better to use class A IP (ex. 127.0.0.1) rather than class C.

src/utils/ip.ts
raosan: Implemented

src/index.ts
raosan: Fixed using mock Ip by adding env check in the function, because i'm stuck adding mock in the test file 😆

src/utils/ip.ts
```diff
@@ -40,3 +44,21 @@ export default function getIp(): string {
 
   return address
 }
+
+export async function getPublicIp() {
+  if (process.env.NODE_ENV === 'test') {
+    publicIpAddress = '127.0.0.1'
+    return
+  }
```
nicnocquee, line 52: it's better to not pollute our code with testing conditionals. Instead, the tests should mock this constant before each tests. For example, in test/index.test.ts

```javascript
// in test/index.test.ts
import sinon from 'sinon'
describe('monika', () => {
  let getPublicIPStub: any
  beforeEach(() => {
    getPublicIPStub = sinon.stub(Util, 'getPublicIp' as never)
    // @ts-ignore
    Util.publicIpAddress = '127.0.01'
  })
  afterEach(() => {
    getPublicIPStub.restore()
  })
```

nicnocquee (changes requested):

src/utils/ip.ts
raosan: fixed. please recheck

nicnocquee (approved):

Issue #313, opened 2021-08-03
Monika wrongly calculates response time when the status is not 2xx


armedi:
**Describe the bug**
When monika gets response from a response with status other than 2xx, it calculates response time as 0ms

**To Reproduce**
1. run monika with this config
```json
{
  "notifications": [
    {
      "id": "desktop",
      "type": "desktop"
    }
  ],
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "method": "GET",
          "url": "https://httpbin.org/status/400",
          "saveBody": false
        }
      ],
      "incidentThreshold": 5,
      "recoveryThreshold": 5
    }
  ]
}
```
2. In the terminal you will see log similar to this:
![Screen Shot 2021-08-03 at 13 38 59](https://user-images.githubusercontent.com/22452017/127969452-eef0f9f5-a730-4aae-8b72-f03c9fa6b74e.png)

The response time is always 0, while it is actually not.

**Expected behavior**
Monika logs correct response time

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS]
 - Version [e.g. 11.1]

PR #327, opened 2021-08-09
Fix response time display error when not 2xx

sapiderman:
# Monika Pull Request 

## What feature/issue does this PR fix
1. Closing #313 

1.  The response time data lives in the config.extraData rather than than in error.response.
2. So returning config.extraData instead when not 2xx
3. Add some documentation and update the sanitize event definition

## How to test  
1. see Issue #313 for how to reproduce

## Screenshot before fix vs after fix
image

### DIFF
--- a/src/components/probe/probing.ts
+++ b/src/components/probe/probing.ts
@@ -66,31 +66,28 @@ export async function probing(
     })
     return res as AxiosResponseWithExtraData
   } catch (error) {
-    let errStatus
+    let errResponseCode
     let errData
     let errHdr
 
     if (error.response) {
       // Axios doesn't always return error response
-      errStatus = error.response.status
+      errResponseCode = error.response.status
       errData = error.response.data
       errHdr = error.response.headers
     } else {
-      errStatus = 500 // TODO: how to detect timeouts?
+      errResponseCode = 500 // TODO: how to detect timeouts?
       errData = ''
       errHdr = ''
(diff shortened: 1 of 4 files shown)


src/components/probe/probing.ts
```diff
@@ -66,31 +66,28 @@ export async function probing(
...
 
     if (error.response) {
       // Axios doesn't always return error response
-      errStatus = error.response.status
+      errResponseCode = error.response.status
       errData = error.response.data
       errHdr = error.response.headers
     } else {
-      errStatus = 500 // TODO: how to detect timeouts?
+      errResponseCode = 500 // TODO: how to detect timeouts?
```
gsukanto, line 79: For detecting timeouts error, we can use:
```
if (error.code === 'ECONNABORTED')
```

gsukanto (approved):
lgtm

src/components/probe/probing.ts
gsukanto: reff: https://github.com/axios/axios/blob/v0.21.1/lib/adapters/http.js#L280

armedi (approved):

src/components/probe/probing.ts
sapiderman: thanks greg.. yup found this

nicnocquee (approved):

Issue #328, opened 2021-08-09
Use native library for sending desktop notifications


dennypradipta:
- [x]  macOS: Use OSAScript (Apple Open Scripting Architecture)
- [x] Linux: Use notify-send command
- [x] Windows: Use Powershell

PR #329, opened 2021-08-09
feat: use native notifications and remove node-notifier

dennypradipta:
# Monika Pull Request (PR)  

1. #328 

1.  Removed node-notifier
2.  Add desktop notification mechanism using native libraries (ported from [https://github.com/ariya/berkala](Berkala))
3.  Update docs

## How to test  
1.  Run monika with desktop notifications
2.  It should send notifications

## Additional context
Please test it in 3 OS: OS X, Linux, and Windows. I already tested in Linux, Kubuntu flavor


src/components/notification/channel/desktop.ts
```diff
@@ -22,12 +22,77 @@
...
+  $notifier = [Windows.UI.Notifications.ToastNotificationManager]::CreateToastNotifier('Monika');
+  $notifier.Show($toast);
+  `
+
+  switch (operatingSystem) {
+    case 'Linux':
+      spawnSync('notify-send', ['-a', 'Monika', title, message])
+      break
+    case 'Darwin':
+      spawnSync('osascript', ['-e', osascript])
```
armedi, line 85: Running on MacOS

Screen Shot 2021-08-09 at 14 57 15


When one of those notification is clicked, this window opened

Screen Shot 2021-08-09 at 14 57 42

armedi: Is it possible to change notification card title from "SCRIPT EDITOR" to "Monika"?

docs/src/pages/guides/notifications.md
```diff
@@ -343,11 +343,11 @@ You can get a notification from Monika to your Whatsapp number without having a
 
 ## Desktop Notifications
 
-Monika uses [node-notifier](https://github.com/mikaelbr/node-notifier#readme) to support desktop notifications. Here are the prerequisites for enabling the desktop notifications:
+Monika supports desktop notifications. Here are the prerequisites for enabling the desktop notifications:
```
gsukanto, line 346: this `supports` suppose to be plural or singular ya?

src/components/notification/channel/desktop.ts
dennypradipta: I updated with title and subtitle, please take a look. I don't have a Mac device so I couldn't test this.

gsukanto (approved):
lgtm

src/components/notification/channel/desktop.ts
nicnocquee: seems like it's not possible to change the "SCRIPT EDITOR" title. So maybe just remove the subtitle. It looks redundant. 
Screenshot 2021-08-10 at 00 00 51

ariya (approved):

src/components/notification/channel/desktop.ts
dennypradipta: We still need the subtitle though, 
image

Title is for the app name, which is Monika
Subtitle is for the event (Monika is Running, New INCIDENT, new RECOVERY)
Message is for message

dennypradipta: I think we should use title for Monika, considering we couldn't remove the "Script Editor" title.

docs/src/pages/guides/notifications.md
dennypradipta: `supports` uses `s` because of the subject (Monika)

PR #342, opened 2021-08-16
refactor: send desktop notification function

armedi:
this refactor is needed to make the `sendDesktop` function more reusable for use cases other than alert notification


src/components/notification/sender.ts
```diff
@@ -202,21 +201,14 @@ export const workplaceNotificationSender = async ({
...
 }) => {
   await sendDesktop({
-    url: data?.url,
-    body: {
-      url: '-',
-      alert: body,
-      time: new Date().toLocaleString(),
-      status,
-    },
+    title: status === 'INIT' ? 'Monika is running' : 'Monika terminated',
```
dennypradipta, line 211: I only read this by code, but if you're gonna do it like this, how are we supposed to get the notifications for the ups and down?

It seems like when there is an alert, it will display "Monika terminated" instead of the alert

dennypradipta (changes requested):

gsukanto (approved):

src/components/notification/sender.ts
armedi: This is one of the thing that we need to refactor. This sender file only for monika start and termination, while for alert is in /notification/index.ts file

sapiderman (approved):
it works!

dennypradipta (approved):

PR #345, opened 2021-08-19
Upload Binary Files when New Release

### LINKED ISSUE #344: Upload pkg binaries when there is a new release

dennypradipta:
# Monika Pull Request (PR)  

This PR resolves #344 

I added a workflow to upload the binary files to the release when a release has been created. It creates 3 binary for Linux, MacOS, and Windows

## How to test  
The only way to test this is by releasing new version. But you could test the generated binary files:

1. Run `npm run pkg -- -t node12-<platform>-x64`. Change `<platform>` with this available options: `linux`, `win`, `macos` 
2. After packaging success, copy your monika.json to dist/ folder. 
3. Change dir to dist/
4. Run your binary with your configuration (You may need to `sudo chmod +x monika` if you're running Linux or MacOS)

## Additional Context

I tested similar workflow in my repo: https://github.com/dennypradipta/discord-bot/actions/runs/1145562768
And the outputs are here: https://github.com/dennypradipta/discord-bot/releases/tag/v0.9

### DIFF
--- a/.github/workflows/upload-binary.yml
+++ b/.github/workflows/upload-binary.yml
@@ -4,9 +4,8 @@
 name: Upload Binary
 
 on:
-  push:
-    tags:
-      - "*"
+  release:
+    types: [published]
 
 jobs:
   test:
--- a/.github/workflows/upload-binary.yml
+++ b/.github/workflows/upload-binary.yml
@@ -1,0 +1,70 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Upload Binary
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npm ci
+        name: Install dependencies
+      - run: npm test
+        name: Run tests
+      - run: npm pack
+        name: Pack using npm pack
+      - run: npm install -g ./hyperjumptech-monika-*.tgz
+        name: Install the packed npm package
+      - run: npm run prod_test
+        name: Run production test
+
+  upload:
+    needs: test
+    name: Publish for ${{ matrix.os }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - os: Linux
+            artifact_name: monika
+            platform: linux-x64
+            file_suffix: linux-x64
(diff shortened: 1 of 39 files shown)


ariya:
> 3. Change dir to dist/

A much safer place to test is to use a completely different directory, e.g. `$HOME/temp`, with no local `.node-modules` around.

package.json
```diff
@@ -28,7 +28,7 @@
     "version": "oclif-dev readme && git add README.md",
     "format": "prettier --write .",
     "check-format": "prettier --check .",
-    "pkg": "npm run prepack && pkg . --compress brotli"
+    "pkg": "npm run prepack && pkg . --compress gzip"
```
ariya, line 31: reason to change to gzip? is it smaller?

dennypradipta: Here are my data for Brotli compression:
Compile time:
image
File size:
image

GZIP:
image
File size:
image

In terms of filesize, there are roughly ~5MB difference. But one thing that should be noted is that the compile time is faster. It probably because of my machine, though. 

As for startup time, there is a slight difference up to ~1 seconds between running using brotli compressed and gzip compressed (Brotli is slower). But I couldn't measure the startup time, so I counted manually

nicnocquee (approved):

package.json
dennypradipta: After some consideration, I changed it back to Brotli. The reason is even though the compile time is longer, it will be compiled only once and users are able to use smaller size binary.

ariya (approved):
:100:

.github/workflows/upload-binary.yml
```diff
@@ -0,0 +1,70 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Upload Binary
+
+on:
```
kevinhermawan, line 6: Why trigger on push tags? Not release created or published?

```
on:
  release:
    types: [created]
```

dennypradipta: I got this reference from https://github.com/ariya/berkala/blob/main/.github/workflows/release.yml

I tested with release: types: [created] too in my personal repo here: https://github.com/dennypradipta/discord-bot/releases/tag/v.0.10 and it works, too.

Which one do you think I should use?

kevinhermawan: I think `on release published` is your purpose. Since the tags can be created without release :)

dennypradipta: Tested in this https://github.com/dennypradipta/discord-bot/releases/tag/v0.11

Cheers!

kevinhermawan (approved):

PR #357, opened 2021-08-27
Added: Secondary proof by checking to STUN server

### LINKED ISSUE #310: Secondary proof
## Problem

There are times when the computer where Monika is running is having connection problem. When this happens, Monika will report and store this as incident in the database even though the probed target probably has no issue. 

## Solution

- [x] Monika should have a separate cycle to check if it can connect to the network. We can check the connection by connecting to a STUN server.
- [x] If the stun server is not reachable, Monika should not probe anything. Then continue again, when the stun is reachable.

raosan:
# Monika Pull Request (PR)  
fix #310 

1. adding secondary proof by checking stun server to check if monika is connected to internet

1.  using stun package that already used to obtain public IP
2. add `isConnectedToSTUNServer` as the status

## How to test  
1. run monika
2. see log that it is connected to stun server and getting public IP
3. disconnect your wifi, and it will log that monika cannot connect to stun server, and no probe is run
4. connect again your wifi, the probe should run as usual
Screen Shot 2021-08-27 at 15 44 08


src/looper.ts
```diff
@@ -121,7 +134,10 @@ function loopProbe(
     if (counter === repeats) {
       clearInterval(probeInterval)
     } else {
-      doProbe(++counter, probe, notifications)
+      // eslint-disable-next-line no-lonely-if
```
gsukanto, line 137: How about using else if instead?

docs/src/pages/guides/cli-options.md
```diff
@@ -144,6 +144,14 @@ You can combine this flag with the `--id` flag to repeat custom sequences.
 monika -r 3 -i 1,3,1
 ```
 
+## STUN
+
+By default monika will continuously checking the STUN server in 20 seconds interval. You can specify the number of interval using `-s` or `--stun` flags followed by a number. For example to set the interval to 10 seconds type the command below:
```
gsukanto, line 149: cosmetic comment: we can add link to explanation what is STUN server.

gsukanto (approved):
lgtm

src/looper.ts
raosan: i think `else if` is better used to compare the same variable. In this case `if (counter === repeats)` and `if (isConnectedToSTUNServer)` doesn't compare same variable, so not a good case to use `else if`. What do you think?

docs/src/pages/guides/cli-options.md
nicnocquee: good catch. add this link https://en.wikipedia.org/wiki/STUN

src/index.ts
```diff
@@ -162,6 +168,12 @@ class Monika extends Command {
       multiple: false,
     }),
 
+    stun: flags.string({
+      char: 's', // (s)stun
+      description: 'interval in seconds to check STUN server',
+      multiple: false,
```
nicnocquee, line 174: put the default '20' here instead of inside `loopCheckSTUNServer`

docs/src/pages/guides/cli-options.md
nicnocquee: also, "You can specify the number of interval using `-s` or `--stun` flags followed by a number in seconds."

src/looper.ts
```diff
@@ -102,6 +103,18 @@ export function isIDValid(config: Config, ids: string): boolean {
   return true
 }
 
+export function loopCheckSTUNServer(interval: number) {
+  const checkSTUNinterval = setInterval(async () => {
```
nicnocquee, line 107: can we also add a condition if interval is < = 0, no need to loop the checkSTUNinterval

docs/src/pages/guides/cli-options.md
nicnocquee: Also mention, that if the number is less than zero, monika will not check the STUN server.

nicnocquee (changes requested):

src/looper.ts
nicnocquee: i think @gsukanto is correct, can just use else if here because there are no other conditionals inside the else block.

src/looper.ts
raosan: no loop, but just check once to get public ip

nicnocquee (approved):

Issue #370, opened 2021-09-03
STUN check loop does not stop on on provided repeat flag


samsunoor:
**Describe the bug**
STUN check looper will still going on even if Monika is provided with repeat flag (`-r`)

**To Reproduce**
1. `npm run start -- -r 1`
2. After done probing, Monika will still run checking STUN server

**Expected behavior**
Monika will stop running after repeat flag number is satisfied.

**Screenshots**
![image](https://user-images.githubusercontent.com/9563873/131949979-70a9892e-0cff-45e1-8d75-1a9298f6c0c2.png)

PR #375, opened 2021-09-06
fix: stop STUN check loop on provided repeat flag

raosan:
# Monika Pull Request (PR)  
fix issue #370 

1. stop looping to check stun server when repeats flag is added

1.  add conditional to check repeats flag on `loopCheckSTUNServer` fuction

## How to test  
1. run `npm run start -- -r 1`
2. see there will be no checking to stun server after 1 time probing
3. if you want monika to also terminating, add status-notification flag `npm run start -- -r 1 --status-notification=false`

### DIFF
--- a/src/looper.ts
+++ b/src/looper.ts
@@ -35,6 +35,8 @@ const MILLISECONDS = 1000
 export const DEFAULT_THRESHOLD = 5
 const DEFAULT_REPORT_INTERVAL = 180000 // 3 minutes
 
+let probeCounter = 0
+
 /**
  * sanitizeProbe sanitize currently mapped probe name, alerts, and threshold
  * @param {object} probe is the probe configuration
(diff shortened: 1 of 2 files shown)


gsukanto (approved):

src/looper.ts
```diff
@@ -35,6 +35,8 @@ const MILLISECONDS = 1000
 export const DEFAULT_THRESHOLD = 5
 const DEFAULT_REPORT_INTERVAL = 180000 // 3 minutes
 
+let probeCounter = 0
```
nicnocquee, line 38: I think this implementation is not right. If I assumed correctly, you extracted the `counter` variable from `loopProbe` function, right? But the problem is, that local `counter` variable is actually corresponded to each of the probes. 

For example, say we have `repeat` flag set to `1`, and we have 2 probes with different intervals, 10 seconds and 20 seconds, respectively. So Monika will actually terminate after 30 seconds, which is after both probes have run once.

So if you make the `counter` variable global like this, only the first probe will run with `repeat` flag set to `1`. This breaks the `repeat` flag feature.

nicnocquee (changes requested):

src/looper.ts
raosan: oh i see...
so the `loopCheckSTUNServer` better has it's own counter then

raosan: fixed. please recheck

src/looper.ts
```diff
@@ -110,19 +110,23 @@ export function isIDValid(config: Config, ids: string): boolean {
...
-      await getPublicIp()
-    }, interval * MILLISECONDS)
+export async function loopCheckSTUNServer(interval: number, repeats: number) {
+  let counter = 0
 
-    if (process.env.CI || process.env.NODE_ENV === 'test') {
+  const checkSTUNinterval = setInterval(async () => {
+    if (counter === repeats) {
+      clearInterval(checkSTUNinterval)
+    } else if (interval <= 0) {
```
nicnocquee, line 119: CMIIW, say we have 1 probe with 5s interval, and the STUN check is 20s interval. So when `repeat` is 1, Monika will still wait for this to execute and finish even though the probe has finished?

Ideally, when `repeat` value exists, all other loop processes should be cancelled when the probes have been repeated x times.

nicnocquee (changes requested):

src/looper.ts
raosan: hmm so i have to check if there is a probe interval aborted, then abort the STUN loop also.

fixed, please recheck

nicnocquee (approved):

PR #388, opened 2021-09-07
Support an object for TLS check domains

DinosaurDad:
TLS domains can now either be a string:
  "foo.com"
or an object with options:
  {"domain": "foo.com", "options": {"path": "/robots.txt"}}

# Monika Pull Request (PR)  

1. Allow a path (or other options) to be passed to TLS checks, because some sites don't allow the root path to be accessed. OpenSSL can verify the domain, but using the `https` class in NodeJS you must provide a path that is valid.

1.  Added a check for an object or a string, and pass an `options` argument to ssl-check

## How to test  
1. Create certificate domains such as this:
```
  "domains": [
      "foo.com",
      {
        "domain": "bar.com",
        "options": {
          "path": "/robots.txt"
        }
      }
  ]
```


DinosaurDad:
I cleaned up the code that deals with domains being string or object. This feels a bit more TypeScript (aka, what was I thinking?)

gsukanto (approved):
lgtm

src/components/tls-checker/index.ts
```diff
@@ -24,15 +24,21 @@
 
 import sslChecker from 'ssl-checker'
 
-export async function checkTLS(url: string, expiryThreshold = 30) {
-  const { valid, validTo, daysRemaining } = await sslChecker(url)
+export async function checkTLS(host: any, expiryThreshold = 30) {
+  const hostOptions = host?.options! ?? {}
+  const domain = host?.domain! ?? host
+
```
nicnocquee, line 30: I think it's easier to understand if we use the exact type for the `host` argument which is either a `string`, or an object with `domain` and `options` properties:

```typescript
type TLSHostArg = {
  domain: string
  options?: Record<string, any>
}
export async function checkTLS(
  host: string | TLSHostArg,
  expiryThreshold = 30
) {
  const hostOptions = (host as TLSHostArg)?.options ?? {}
  const domain = (host as TLSHostArg)?.domain ?? (host as string)
```

src/components/tls-checker/index.ts
```diff
@@ -24,15 +24,21 @@
 
 import sslChecker from 'ssl-checker'
```
nicnocquee, line 25: - [ ] please also add a test in `src/components/tls-checker/index.test.ts`
- [ ] please update the documentation in `docs/src/pages/guides/tls-checkers.md`

nicnocquee (changes requested):

src/components/tls-checker/index.ts
DinosaurDad: I'm not seasoned with TypeScript, so I'm not familiar with the ramifications of that. If that would work here, go for it.

DinosaurDad: Done

src/components/tls-checker/index.ts
DinosaurDad: Added and updated.

nicnocquee (approved):

PR #390, opened 2021-09-08
Added: alert for individual request

### LINKED ISSUE #373: Add ability to refer to specific request in alert query
Currently in the alert configuration, user can query for response.

For example:
```json
{
  "query": "response.body.data.id != 2",
  "subject": "Response data check",
  "message": "User id is {{ response.body.data.id }}, expecting 2"
}
```

This alert will run for every request.

If there are multiple requests, that will cause the probe to always fail.

So it should be possible to run a check only for specific request.

For example:
```json
{
  "query": "responses[0].body.data.id != 2",
  "subject": "Response data check",
  "message": "User id is {{ responses[0].body.data.id }}, expecting 2"
}
```

armedi:
# Monika Pull Request (PR)  

This PR resolves #373 

1.  added option to configure alert for one alert, user can do this by adding `alerts` property to request configuration object.

## How to test  
1. run monika with config below
```json
{
  "notifications": [
    {
      "id": "desktop",
      "type": "desktop"
    }
  ],
  "probes": [
    {
      "id": "httpbin",
      "name": "probe httpbin",
      "requests": [
        {
          "method": "GET",
          "url": "https://httpbin.org/status/200"
        },
        {
          "method": "GET",
          "url": "https://httpbin.org/status/400",
          "alerts": [
            {
              "query": "response.status != 200",
              "message": "HTTP status is {{ response.status }}, expecting 200"
            }
          ]
        }
      ],
      "alerts": [
        {
          "query": "response.time > 1000",
          "message": "Response time is {{ response.time }}ms, expecting less than 1000ms"
        }
      ],
      "incidentThreshold": 1,
      "recoveryThreshold": 1
    }
  ]
}
```


gsukanto (approved):

nicnocquee (commented):
Is there no tests for this new behaviour?

armedi:
> Is there no tests for this new behaviour?

I need advise on how to test this behavior other than hitting real http endpoint and check the stdout

sapiderman (approved):
this works. nice!

PR #395, opened 2021-09-09
Split overriding notification flag

### LINKED ISSUE #383: Separate Notification File that will overrides Monika configuration
I think it would be nice to have an option to load separate notification file.

Let's say that you share a configuration with your co-workers, like so:
```
{
  "notifications": [
    {
      "id": "unique-id-desktop",
      "type": "desktop"
    }
  ],
  "probes": [
    {
      "requests": [
        {
          "url": "https://github.com"
        }
      ]
    }
  ]
}
```

By default, it will show you the desktop notifications. If we can separate the notifications file and pass it through the CLI options, like so:
`monika -c monika.json --notification-file notif-config.json`

The notif-config.json will override the monika.json notifications block. So that you can use default notification from the configuration if there is no --notification-flag, but overrides the default if you supply the --notification-flag.

samsunoor:
# Monika Pull Request (PR)  
This PR resolves #383 

1. Add `--notification` flag
2. This flag will override the notifications provided by the `-c/--config` flag

## How to test  
1. `npm start -- --notification path/to/notifications/array.json`


monika-notifications.json
```diff
@@ -0,0 +1,28 @@
+{
+  "notifications": [
+    {
+      "id": "monika-webhook",
+      "type": "monika-notif",
+      "data": {
+        "url": "https://whatsapp.hyperjump.tech/api/notify?token=ce0adb84-c3ce-4590-b3a9-1f30d0d24eb5"
```
nicnocquee, line 7: remove or redact the token please

src/index.ts
```diff
@@ -181,12 +181,18 @@ class Monika extends Command {
...
 
     'status-notification': flags.string({
-      description: 'cron syntax for status notification schedule',
+      description: 'Cron syntax for status notification schedule',
+    }),
+
+    notification: flags.string({
+      description:
+        'Override notification(s) provided by config (-c/--config) with a separate file',
+      dependsOn: ['config'],
```
nicnocquee, line 195: CMIIW, with `dependsOn: ['config']` here, user won't be able to execute `monika --notification <path to notif>` and use `monika.json` file in the current directory right?

If it is as I thought, no need to use dependsOn here.

src/components/config/parse.ts
```diff
@@ -48,7 +53,11 @@ export const parseConfig = (configPath: string, type: string): Config => {
       return (cfg as unknown) as Config
     }
 
-    return JSON.parse(configString)
+    if (omitNotifications) {
+      return JSON.parse(configString) as Omit<Config, 'notifications'>
+    }
+
+    return JSON.parse(configString) as Config
```
nicnocquee, line 60: This line 57 and 60 actually return the same thing. You just change the types but the values itself in runtime are the same. To prove, see [this sample in this typescript playground](https://www.typescriptlang.org/play?#code/ATBQEsDsBcFMCcBmBDAxrYBhA9pR4BzYAb1BBEmQFtYAuYAZ2nigIG4zzkC7hIBXKgCMEHAL6hQqXE2AATZNGQBlZq2ABeYAANQpAJAAiSjUP1DyQwBpQR7rDPAAjKAm6pM6PMXJNwAFLKAPIAcgB0AA7I8AywABQKSqoskAQAlMDIDFi4+AQekAzYADawYcXYBHHEichiaZLShV61AEx+gaGR0bEJPsmsGVnAQVTg0AA8OHiEVsAA5PbzAHwFRaXlldVt9UA). Press run then check the log.

If you actually want to literally remove the `notifications` property, you need to something like

```
const obj = JSON.parse(configString)
delete obj.notifications
```

src/components/config/index.ts
```diff
@@ -126,9 +127,15 @@ export const setupConfigFromFile = async (flags: any, watch: boolean) => {
...
+  checkingInterval: number,
+  omitNotifications: boolean
 ) => {
-  const fetched = await fetchConfig(url)
+  let fetched: Config
+  if (omitNotifications) {
+    fetched = (await fetchConfig(url)) as Omit<Config, 'notifications'>
+  } else {
+    fetched = await fetchConfig(url)
+  }
```
nicnocquee, line 138: same comment with that in parse.ts file.

src/components/config/parse.ts
```diff
@@ -28,8 +28,13 @@ import { parseConfigFromPostman } from './parse-postman'
 import { parseHarFile } from './parse-har'
 import path from 'path'
 import yml from 'js-yaml'
+import { Notification } from '../../interfaces/notification'
 
-export const parseConfig = (configPath: string, type: string): Config => {
+export const parseConfig = (
+  configPath: string,
+  type: string,
+  omitNotifications: boolean
```
nicnocquee, line 36: it's better to make this optional, so we don't need existing callers' signatures.

nicnocquee (changes requested):

ariya:
How about making this more generic by allowing configuration merge across multiple files? IOW, with this:

```bash
monika -c config1.json config2.json
```

Monika simply slurps the probes and notifications (if any) from `config1.json`, continues with `config2.json`, and overrides what were specified before.

Effectively, this means that `config2.json` can be about notifications, if e.g. `config1.json` does not specify any.

src/index.ts
samsunoor: I tested this and the dependency declaration does not affect that `--config`'s default behaviour

samsunoor:
> How about making this more generic by allowing configuration merge across multiple files? IOW, with this:
> 
> ```shell
> monika -c config1.json config2.json
> ```
> 
> Monika simply slurps the probes and notifications (if any) from `config1.json`, continues with `config2.json`, and overrides what were specified before.
> 
> Effectively, this means that `config2.json` can be about notifications, if e.g. `config1.json` does not specify any.

Should we support both merge and override behaviour?
e.g. multiple arguments to `--config config.json config2.json` will merge them,
but `--config config.json config2.json --override true` will make the later config to override previous config?

nicnocquee:
> > How about making this more generic by allowing configuration merge across multiple files? IOW, with this:
> > ```shell
> > monika -c config1.json config2.json
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Monika simply slurps the probes and notifications (if any) from `config1.json`, continues with `config2.json`, and overrides what were specified before.
> > Effectively, this means that `config2.json` can be about notifications, if e.g. `config1.json` does not specify any.
> 
> Should we support both merge and override behaviour?
> e.g. multiple arguments to `--config config.json config2.json` will merge them,
> but `--config config.json config2.json --override true` will make the later config to override previous config?

Oh @ariya's suggestion makes sense. So no need for `--notification` or `--override true`.

gsukanto (approved):

ariya:
> e.g. multiple arguments to `--config config.json config2.json` will merge them,

TBH, I can't think of a real use case for a true config merge, as I believe config overriding (the latter "replaces" the former) is what most people want.

samsunoor:
move to #407

PR #407, opened 2021-09-15
Add support multiple args to config flag

### LINKED ISSUE #383: Separate Notification File that will overrides Monika configuration

samsunoor:
# Monika Pull Request (PR)  
This resolves #383 

1. Add support for `--config` to accept multiple sources of configurations

1.  Later configurations will replace previous top-level keys

## How to test  
1. `npm run start -- -c monika.example.json monika.example.yml` will take `example.com` as its probe


docs/src/pages/guides/cli-options.md
```diff
@@ -26,7 +26,19 @@ Configuration files may be placed remotely which you can specify using the same
 monika -c https://raw.githubusercontent.com/hyperjumptech/monika/main/config_sample/config.desktop.example.yml
 ```
 
-A neat feature is that the configuration file will be re-read and monitoring updated if Monika detects any changes to it.
+A neat feature is that the configuration file will watch any changes to the file and update running config on-the-fly.
```
nicnocquee, line 29: A neat feature is that the configuration file is watched and any changes will cause Monika to reload.

src/interfaces/config.ts
```diff
@@ -36,3 +36,16 @@ export interface Config {
...
+ */
+export interface ConfigOptional {
+  certificate?: Certificate
+  interval?: number
+  notifications?: Notification[]
+  probes?: Probe[]
+  'status-notification'?: string
+  symon?: SymonConfig
+  version?: string
+}
```
nicnocquee, line 51: Is it not possible to use `Partial<Config>` instead? so no need to define another interface. Check this [SO](https://stackoverflow.com/a/40076355/401544).

src/components/config/index.ts
```diff
@@ -85,6 +74,103 @@ const handshakeAndValidate = async (config: Config) => {
...
+      probes:
+        current?.probes && current.probes.length > 0
+          ? current.probes
+          : prev.probes,
+      symon: current?.symon ? current.symon : prev.symon,
+      'status-notification': current?.['status-notification']
+        ? current['status-notification']
+        : prev['status-notification'],
+    }
+  })
```
nicnocquee, line 108: Since the next config will override the previous one, can we not just use

```
configs.reduce((prev, curr) => {
   return {
      ...prev,
      ...curr
   }
})
```

nicnocquee (changes requested):

gsukanto (approved):

nicnocquee (approved):

PR #410, opened 2021-09-16
Changed: Attach incident and recovery thresholds calculation to each request and per alert

### LINKED ISSUE #403: Incident and recovery threshold calculation
Right now, the thresholds are attached to probe. For example when monika runs with this config

```json
{
  "probes": [
    {
      "requests": [
        {
          "method": "GET",
          "url": "https://example.com/1"
        },
        {
          "method": "GET",
          "url": "https://example.com/2"
        }
      ],
      "alerts": [
        {
          "query": "response.time > 1000",
          "message": "Response time is {{ response.time }}ms, expecting less than 1000ms"
        },
        {
          "query": "response.status != 200",
          "message": "Response status is {{ response.status }}, expecting 200"
        }
      ],
      "incidentThreshold": 5,
      "recoveryThreshold": 5
    }
  ]
}
``` 

- when the `response.time > 1000` alert is triggered for 4 times in a row, then in the fifth iteration of the request the triggered alert is `response.status != 200`, monika will send notification about the incident of `response.status != 200` only.
- when alerts are triggered 4 times in a row for the first request, then in the fifth iteration it's the second request that triggers the alert,  monika will send a notification about the incident of the second request only.

all of the behaviors above are because threshold calculation is attached to probe, no matter what alert happens and in which request it happens.

I think it is better to attach the threshold calculation to each alert and per request. So with configuration above, monika should calculate each alert incidents separately and per request.

- Initial counter values
   - firstRequestConsecutiveResponseTimeIncident = 0
   - firstRequestConsecutiveResponseStatusIncident = 0
   - secondRequestConsecutiveResponseTimeIncident = 0
   - secondRequestConsecutiveResponseStatusIncident = 0

- the `response.time > 1000` alert is triggered for 4 times in a row for first request
   - firstRequestConsecutiveResponseTimeIncident = 4
   - firstRequestConsecutiveResponseStatusIncident = 0
   - secondRequestConsecutiveResponseTimeIncident = 0
   - secondRequestConsecutiveResponseStatusIncident = 0

- in the fifth iteration the first request no longer trigger `response.time > 1000` alert, but it triggers `response.status != 200` alert
   - firstRequestConsecutiveResponseTimeIncident = 0
   - firstRequestConsecutiveResponseStatusIncident = 1
   - secondRequestConsecutiveResponseTimeIncident = 0
   - secondRequestConsecutiveResponseStatusIncident = 0
   
   no alert reaches the threshold yet, and notification sent

- in the next iteration the first request does not trigger alert, but the second request trigger `response.status != 200` alert
   - firstRequestConsecutiveResponseTimeIncident = 0
   - firstRequestConsecutiveResponseStatusIncident = 0
   - secondRequestConsecutiveResponseTimeIncident = 0
   - secondRequestConsecutiveResponseStatusIncident = 1

So, the notification should only be sent when a request triggers an alert multiple times in a row and it reaches the threshold value. All calculations should be done per request and per alert.

The same for recovery calculation

armedi:
# Monika Pull Request (PR)  

This PR resolves #403 

1. Move threshold calculation to after each request's response is received
2. calculate threshold for each request and per alert

## Additional note
1. I completely rewrote src/components/notification/process-server-status.ts file. Now server state calculation is using xstate instead of using multiple ifs for better readability

## How to test  
1. npm test
2. run monika with this config
```yaml
notifications:
  - id: desktop
    type: desktop
probes:
  - id: httpbin
    name: probe httpbin
    requests:
      - method: GET
        url: https://httpbin.org/status/400
      - method: GET
        url: https://httpbin.org/status/500
        alerts:
          - query: response.status != 200
            message: HTTP status is {{ response.status }}, expecting 200
    alerts:
      - query: response.time > 1200
        message: Response time is {{ response.time }}ms, expecting less than 1200ms
    incidentThreshold: 2
    recoveryThreshold: 2

```

### DIFF
--- a/src/components/notification/process-server-status.ts
+++ b/src/components/notification/process-server-status.ts
@@ -25,9 +25,9 @@
 import { setAlert } from '../../components/logger'
 import { LogObject } from '../../interfaces/logs'
 import { Probe } from '../../interfaces/probe'
-import { ProbeStatus, ProbeStateDetails } from '../../interfaces/probe-status'
+import { ProbeStateDetails, ProbeStatus } from '../../interfaces/probe-status'
 import { AxiosResponseWithExtraData } from '../../interfaces/request'
-import { ValidateResponse } from '../../plugins/validate-response'
+import { ValidatedResponse } from '../../plugins/validate-response'
 import { log } from '../../utils/pino'
 
 let PROBE_STATUSES: ProbeStatus[] = []
@@ -64,7 +64,7 @@ const determineProbeState = ({
 }: {
   errorName: string
   probeStatusDetail: ProbeStateDetails
-  validation: ValidateResponse
+  validation: ValidatedResponse
   incidentThreshold: number
   recoveryThreshold: number
 }) => {
@@ -190,7 +190,7 @@ export const processThresholds = ({
   probe: Probe
   probeRes: AxiosResponseWithExtraData
   totalRequests: number
-  validatedResp: ValidateResponse[]
+  validatedResp: ValidatedResponse[]
   mLog: LogObject
 }) => {
   try {
--- a/src/components/notification/process-server-status.ts
+++ b/src/components/notification/process-server-status.ts
@@ -27,7 +27,6 @@ import { Probe } from '../../interfaces/probe'
 import { ServerAlertState } from '../../interfaces/probe-status'
 import { RequestConfig } from '../../interfaces/request'
 import { ValidatedResponse } from '../../plugins/validate-response'
-import { log } from '../../utils/pino'
 
 type ServerAlertStateContext = {
   incidentThreshold: number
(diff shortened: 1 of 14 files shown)


gsukanto (approved):

sapiderman (approved):

src/components/notification/process-server-status.ts
```diff
@@ -22,262 +22,155 @@
...
+              actions: 'handleFailure',
+            },
+            {
+              actions: 'handleFailure',
+            },
+          ],
+          SUCCESS: {
+            actions: 'handleSuccess',
+          },
+        },
```
nicnocquee, line 68: ok, xstate is cool! 👍

nicnocquee (approved):

PR #425, opened 2021-09-19
Add Larksuite notification

sapiderman:
# Monika Pull Request (PR)  

1. Add notifcation for Larksuite
2. Add some documentation and sample configuration

1.  Add new template for Lark messages via webhook

## How to test  
1. Sign up to larksuite and create a team
2. Create a webhook and use it as notification config
3. Run Monika

Monika config may be: 
```yml
notifications:
  - id: lark-group-message
    type: lark
    data:
      url: https://open.larksuite.com/open-apis/bot/v2/hook/larksuit-url-here
```

## Screenshot 
image


src/components/notification/channel/lark.ts
```diff
@@ -0,0 +1,159 @@
...
+      headers: {
+        Accept: 'application/json',
+        'Content-type': 'application/json',
+      },
+      data: larkMessage,
+    })
+
+    return res
+  } catch (error) {
+    throw error
```
gsukanto, line 157: If in the end we will throw the same error, why do we need to have try/catch in here?
It would be simpler to not use try/catch instead.

gsukanto (approved):

src/components/notification/channel/lark.ts
sapiderman: thanks, great feedback. updated.

nicnocquee (approved):

Issue #416, opened 2021-09-16
Running average, max, and min response time in status notification


nicnocquee:
Since Monika will have database-less mode (#415), the average response time sent in the periodic status notification should not depend on the records in the database. So as per our discussion, the average response time should be calculated every time a new response is received.

- [x] calculate running average response time
- [x] calculate running minimum response time
- [x] calculate running maximum response time

PR #442, opened 2021-09-27
Fixed: Running average, max, and min response time in status notification

raosan:
# Monika Pull Request (PR)  
fix issue #416 

1. 

1.  

## How to test  
1. 
2. 
3.

### DIFF
--- a/src/components/probe/index.ts
+++ b/src/components/probe/index.ts
@@ -242,5 +242,5 @@ function calculateResponseTime(probeRes: ProbeRequestResponse) {
     minResponseTime = probeRes.responseTime
   }
 
-  averageResponseTime = maxResponseTime + minResponseTime / 2
+  averageResponseTime = (maxResponseTime + minResponseTime) / 2
 }
--- a/src/components/probe/index.ts
+++ b/src/components/probe/index.ts
@@ -52,6 +52,11 @@ interface ProbeSendNotification extends Omit<ProbeStatusProcessed, 'statuses'> {
   probeState?: ServerAlertState
 }
 
+// global variable
+let maxResponseTime = 0
+let minResponseTime = 0
+export let averageResponseTime = 0
+
 // Probes Thresholds processed, Send out notifications/alerts.
 async function checkThresholdsAndSendAlert(
   data: ProbeStatusProcessed,
@@ -143,6 +148,8 @@ export async function doProbe(
       // eslint-disable-next-line no-await-in-loop
       const probeRes: ProbeRequestResponse = await probing(request, responses)
 
+      calculateResponseTime(probeRes)
+
       eventEmitter.emit(events.probe.response.received, {
         probe,
         requestIndex,
@@ -221,4 +228,19 @@ export async function doProbe(
       }
     }
   }
+}
+
+function calculateResponseTime(probeRes: ProbeRequestResponse) {
+  if (maxResponseTime === 0 && minResponseTime === 0) {
+    // first time
+    maxResponseTime = probeRes.responseTime
+    minResponseTime = probeRes.responseTime
+  } else if (probeRes.responseTime > maxResponseTime) {
+    maxResponseTime = probeRes.responseTime
+  } else {
+    // probeRes.responseTime < minResponseTime
+    minResponseTime = probeRes.responseTime
+  }
+
+  averageResponseTime = maxResponseTime + minResponseTime / 2
 }
--- a/src/jobs/summary-notification.ts
+++ b/src/jobs/summary-notification.ts
@@ -32,6 +32,7 @@ import { getContext } from '../context'
 import getIp from '../utils/ip'
 import { log } from '../utils/pino'
 import { publicIpAddress } from '../utils/public-ip'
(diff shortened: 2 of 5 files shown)


src/components/probe/index.ts
```diff
@@ -222,3 +231,20 @@ export async function doProbe(
...
+    minResponseTime = probeRes.responseTime
+  } else if (probeRes.responseTime > maxResponseTime) {
+    maxResponseTime = probeRes.responseTime
+  } else if (probeRes.responseTime < minResponseTime) {
+    minResponseTime = probeRes.responseTime
+  }
+
+  totalResponseTime += probeRes.responseTime
+  averageResponseTime = Math.floor(totalResponseTime / responseCount)
+}
```
nicnocquee, line 250: 1.  can we separate this into different file for the sake of separation of concern?
2. It should calculate max, min, and ave in the last 24 hours. Maybe add another variable to store a starting timestamp, then reset the max,min,and ave values in this function when current - starting is 24 hours.

src/jobs/summary-notification.ts
```diff
@@ -48,7 +53,9 @@ export async function getSummaryAndSendNotif() {
       body: `Status Update ${format(new Date(), 'yyyy-MM-dd HH:mm:ss XXX')}
 Host: ${hostname()} (${[publicIpAddress, getIp()].filter(Boolean).join('/')})
 Number of probes: ${summary.numberOfProbes}
-Average response time: ${summary.averageResponseTime} ms in the last 24 hours
+Maximum response time: ${maxResponseTime} ms in the run time
+Minimum response time: ${minResponseTime} ms in the run time
+Average response time: ${averageResponseTime} ms in the run time
```
nicnocquee, line 58: It should say in the last 24 hours.

nicnocquee (changes requested):

src/components/probe/index.ts
```diff
@@ -52,6 +52,13 @@ interface ProbeSendNotification extends Omit<ProbeStatusProcessed, 'statuses'> {
   probeState?: ServerAlertState
 }
 
+// global variable
+let responseCount = 0
```
gsukanto, line 56: Having global variable in here can lead to race condition/mutex.
It would be hard for other developer(s) to understand what will happen if we are running multiple probe at the same time. Will it cause mutex/race condition problem or not.
Can we have comment (description) in here: why we choose to use global variable and what will happen to this global variable if we run multiple probe?

gsukanto (approved):

src/components/probe/index.ts
raosan: we can say it is safe to use global variable here since the probe is processed sequentially as mentioned in probe/index.ts line 143: intentionally wait for a request to finish before processing next request in loop

i have tested it, and i have added it to the comment too

src/components/logger/response-time-log.ts
```diff
@@ -0,0 +1,75 @@
...
+    return true
+  }
+
+  return false
+}
+
+export function logResponseTime(probeRes: ProbeRequestResponse) {
+  if (checkIs24HourHasPassed()) {
+    resetlogs()
+  }
```
armedi, line 59: Maybe adding to @nicnocquee comment https://github.com/hyperjumptech/monika/pull/442#discussion_r716997907

The default for status-notification schedule is 06:00 AM

For instance I start monika on Monday 05:30 AM, then 30 minutes later I will receive summary from 05:30 to 06:00 AM. That's fine for the first day.

The next day Tuesday at 05:30 logs will be reset. Then 30 minutes later monika sends status notification. But now what I receive is only summary from 05:30 to 06:00, not the last 24 hours.

raosan: ah nice catch. So it's better to reset logs after sending the notification, no need checking the 24 hour

armedi: Resetting the logs after sending the notification also lead to unexpected behavior.

For example when you set monika to send notification every 30 minutes

```sh
monika --status-notification "*/30 * * * *"
```

resetting the logs after sending notification will cause the summary to be valid only for the last 30 minutes, not 24 hours.

raosan: I see. I'll have to think about it first

raosan: fixed, please recheck

src/components/notification/channel/lark.ts
```diff
@@ -118,7 +118,15 @@ export const sendLark = async (
                 },
                 {
                   tag: 'text',
-                  value: `${message.meta.averageResponseTime} ms in the last 24 hours`,
+                  value: `${message.meta.maxResponseTime} ms in the run time`,
```
nicnocquee: instead of `in the run time`, can we get the value of duration since the last status notification? if not, better to say `ms since the last status report.`

nicnocquee (changes requested):

src/components/notification/channel/lark.ts
raosan: that's outdated code. i have changed it to in the last X hours. with minimum is 1 hour

PR #444, opened 2021-09-28
Feature: limit database

### LINKED ISSUE #414: Limit the database size
- [ ] By default, the log database should be limited to 1GB. If it grows more than the limit, delete the oldest requests entries.
- [ ] The limit should be configurable in the yml config: `max-db-size: 1GB` or something.

ilmiawan:
# Monika Pull Request (PR)  
fix #414 

Need to limit database size

Add cron job for running tasks to delete data from db

## How to test  
1. set database limit to your configuration
```
db_limit:
  max_db_size: 1000000000
  deleted_data: 1
  cron_schedule: '*/1 * * * *'
```
2. run the application
3. check that the cron job should run and delete 1 data every minute


gsukanto (approved):

src/components/logger/history.ts
```diff
@@ -98,6 +98,33 @@ export async function openLogfile() {
   }
 }
 
+export async function deleteFromProbeRequests(limit: number) {
+  const getIdsToBeDeleted = `SELECT id FROM probe_requests order by created_at asc limit ${limit}`
```
nicnocquee, line 102: so if the limit is 1, which is the default, this will only delete 1 request? is that enough to keep the db size under the max_db_size?

src/loaders/jobs.ts
```diff
@@ -24,8 +24,13 @@
...
 import { tlsChecker } from '../jobs/tls-check'
+import { check_db_size } from '../jobs/check-database'
+import { getConfig } from '../components/config'
 
 export function jobsLoader() {
+  const { db_limit } = getConfig()
   // schedule TLS checker every day at 00:00
   cron.schedule('0 0 * * *', tlsChecker)
+  // schedule database size check
+  cron.schedule(db_limit.cron_schedule, check_db_size)
```
dennypradipta, line 35: Do we need to check if db_limit.cron_schedule is exists to run this command?

src/components/logger/history.ts
ilmiawan: done, by keep deleting untill file size is reached.

src/loaders/jobs.ts
ilmiawan: done

src/jobs/check-database.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  if (!db_limit.deleted_data) {
+    return
+  }
+
+  if (stats.size > db_limit.max_db_size) {
+    const probe_res = await deleteFromProbeRequests(db_limit.deleted_data)
+    const notif_res = await deleteFromNotifications(db_limit.deleted_data)
+    const alert_res = await deleteFromAlerts(db_limit.deleted_data)
```
armedi, line 56: I think there will be problem with this approach of deleting 1 row from probeRequest, notification, and alert in each cycle when user choose to keep logs for all request (`monika --keep-verbose-logs`):

**The amount of data for probeRequest will far exceed alert and notification**

| time | probeRequest | alert | notification |
| ---: | :----------: | :---: | :----------: |
|    1 |      v       |       |              |
|    2 |      v       |       |              |
|    3 |      v       |   v   |              |
|    4 |      v       |   v   |              |
|    5 |      v       |       |              |
|    6 |      v       |       |              |
|    7 |      v       |   v   |              |
|    8 |      v       |       |              |
|    9 |      v       |   v   |              |
|   10 |      v       |   v   |              |
|   11 |      v       |       |              |
|   12 |      v       |   v   |              |
|   13 |      v       |   v   |              |
|   14 |      v       |   v   |      v       |
|   15 |      v       |       |              |
|   16 |      v       |       |              |
|   17 |      v       |       |      v       |
|   18 |      v       |   v   |              |
|   19 |      v       |       |              |
|   20 |      v       |       |              |

For example, the table above illustrate the record of data and the time when it happens. When we remove 3 record for each table then 
- the probeRequest table is deleted until time 18
- the alert table is deleted until time 13
- the notification table is now empty

I think the better approach is to delete records of all table up to certain time, in the case above maybe up to time 17

ilmiawan: done, delete alert and notifications based on probe_requests.

nicnocquee (approved):

Issue #494, opened 2021-10-19
Sending Notification to Google Chat


ariawan:
add how to in docs to sending notification from monika to google chat based on this https://developers.google.com/chat/how-tos/webhooks

PR #523, opened 2021-11-08
Add Google chat notification channel

sapiderman:
# Monika Pull Request (PR)  

1. Closes issue #494 
2. Adds google chat as a notification channel

1.  Add handling for type 'google-chat'
2. Add  json formatting specific for google chat.

## How to test  
1. Create a new notification of type 'google-chat'
2. Add the spaces/google chat bot and generate the url webhook like in the google instruction
3. Run monika and observe notifications on Google Chat space

startup message
image

incident message
image


src/components/notification/channel/googlechat.ts
```diff
@@ -0,0 +1,233 @@
...
+export const sendGoogleChat = async (
+  data: GoogleChatData,
+  message: NotificationMessage
+) => {
+  const notifType =
+    message.meta.type[0].toUpperCase() + message.meta.type.substring(1)
+  let chatMessage
+
+  switch (message.meta.type) {
+    case 'start':
```
dennypradipta, line 39: I think for the "start" case, the `subtitle` should be "Monika is starting" just like the Teams version.
image


And for the sections, it only prints the message body. Time is not needed as it is already at the top:
image

src/components/notification/channel/googlechat.ts
```diff
@@ -0,0 +1,233 @@
...
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      }
+      break
+    case 'termination':
+    case 'incident':
```
dennypradipta, line 74: For both incidents and recoveries, because Google Chat doesn't render the sections with indents like Microsoft Teams:
image

In the google chat:
image

I think it's best to add a colon after the field name such as:
```
Message: ${message.summary}
URL: ${message.meta.url} 
```
and so on.

Or another option, give a line break:
```
Message:
${message.summary}

URL:
${message.meta.url}
```
and so on.

dennypradipta (changes requested):

dennypradipta:
Also please check why the tests is failing :grinning:

gsukanto (approved):

src/components/notification/channel/googlechat.ts
sapiderman: image
adding colons

sapiderman: image

dennypradipta (approved):

PR #539, opened 2021-12-09
Feat: Articles in Docs Sidebar

dennypradipta:
# Monika Pull Request (PR)  

1. This PR solves #535 

1.  Use Medium RSS feeds
2. Create a simple API to fetch the RSS and process it
3. Display processed data to React

## How to test  
1. `npm run dev`
2. Go to <localhost:3000/articles>

## Screenshots
image

## Additional Context
I couldn't seem to get the images because the RSS feeds has no `image` or `media` field.


docs/src/manifests/manifest.json
```diff
@@ -13,6 +13,11 @@
           "title": "Quick Start",
           "path": "/quick-start",
           "editUrl": "/quick-start.md"
+        },
+        {
+          "title": "Articles",
```
sapiderman, line 18: "Articles" terlalu generic gk? Mungkin bisa lebih deskriptif? How-Tos? Self Help Corner? Knowledge base?

sapiderman (approved):
1 comment... Otherwise all good!

docs/src/manifests/manifest.json
dennypradipta: Need to confirm with @nicnocquee first :wink:

nicnocquee: Self help corner sounds like monika is too depressing to use that we need to have a section called that. 😅 Knowledge base usually for FAQs I think. I think Articles is ok for now.

nicnocquee (approved):

PR #573, opened 2022-01-13
Fix [552]:  memory leak on symon mode

ilmiawan:
# Monika Pull Request (PR)  

There are memory leaks when monika is run in symon-mode and no connection to the symon

1. do not send report if there is no connection to symon 
2. clear interval for reporting to symon when there is no connection
3. set new interval for sending report when symon is re connected

## How to test  
1. install node clinic js : `npm i -g clinic`
2. run monika in symon mode using clinic : `clinic doctor -- node bin/run --symonUrl https://symon.hyperjump.tech --symonKey YOUR_KEY_HERE --verbose`
3. let monika run for 2 minutes
4. kill connection
5. let monika run for another 3 minutes
6. stop monika by using `ctrl + c`, wait untill the clinic js result is showing on your default browser 
7. It should show something like in the below.

Screen Shot 2022-01-13 at 16 40 13


nicnocquee (changes requested):
I think we need to add one more listener, something like `onSymonConnectionChange` in `SymonClient`. So when `hasConnectionToSymon` changes to false, Monika can pause all the probing, and resume when `hasConnectionToSymon` back to true.

sapiderman (approved):
Awesome! works for me! 👍🏽 🥇

ilmiawan:
> I think we need to add one more listener, something like `onSymonConnectionChange` in `SymonClient`. So when `hasConnectionToSymon` changes to false, Monika can pause all the probing, and resume when `hasConnectionToSymon` back to true.

Done this. please check again @nicnocquee

src/looper.ts
```diff
@@ -185,8 +190,15 @@ export function idFeeder(
...
   }
 
   return abort
 }
+
+/**
+ * clearProbeInterval clear all probing process
+ */
+export function clearProbeInterval() {
+  clearInterval(probeInterval)
```
nicnocquee, line 203: this doesn't actually clear all intervals because probeInterval is just one interval. to clear all intervals, you should actually use the code in line 193.

nicnocquee (changes requested):

src/looper.ts
ilmiawan: updated

nicnocquee (approved):

Issue #551, opened 2021-12-28
Ping (ICMP) support


nicnocquee:
It'd be great if Monika can also ping a host. We can add a new `ping` property in the `request` 

Example,

```yaml
probes:
  - id: probe-1
    name: Probe Name
    description: Landing page of Probe
    interval: 20
    requests:
      - host: https://hyperjump.tech
         ping: true
```

nicnocquee:
we can use https://www.npmjs.com/package/icmp

PR #579, opened 2022-01-21
Add ping feature

sapiderman:
## Monika Pull Request (PR)  

1. Implements ping as a request
2. Closing #551

## Implementation 

1. Add documentation for ping
2. Add ping as a type of probe request
```yaml
probes:
  - id: 'ping_test'
    name: ping_test
    description: requesting icmp ping
    interval: 10
    requests:
      - url: http://google.com  
      ping: true
```
3. Update validation to account for ping host not need http/https, and filtering for it.
4. Add alert if ping is not alive.
5. Update logger to map ping result properly via our standard logger/output.

image

### DIFF
--- a/src/components/config/validate.ts
+++ b/src/components/config/validate.ts
@@ -34,7 +34,7 @@ import { compileExpression } from '../../utils/expression-parser'
 import type { SymonConfig } from '../reporter'
 import { newPagerDuty } from '../notification/channel/pagerduty'
 
-const HTTPMethods = new Set([
+const HTTPMethods = [
   'DELETE',
   'GET',
   'HEAD',
@@ -46,7 +46,7 @@ const HTTPMethods = new Set([
   'LINK',
   'UNLINK',
   'PING', // fake method for ping
-])
+]
 
 const setInvalidResponse = (message: string): Validation => ({
   valid: false,
@@ -299,13 +299,11 @@ export const validateConfig = (configuration: Config): Validation => {
         }
       }
 
-      if (request.ping === true) {
-        request.method = 'PING' // ok PING is not an HTTP method, but for the purposes of logging, filtering and display
-      } else if (!request.method) {
+      if (!request.method) {
         request.method = 'GET'
       }
 
-      if (!HTTPMethods.has(request.method.toUpperCase()))
+      if (HTTPMethods.indexOf(request.method.toUpperCase()) < 0)
         return PROBE_REQUEST_INVALID_METHOD
     }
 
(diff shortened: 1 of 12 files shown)


src/utils/ping.ts
```diff
@@ -22,14 +22,18 @@
...
 /**
  * sendPing() sends a ping to an address/host
  * @param host is a string
- * @returns
+ * @returns result of ping.promise.probe()
  */
 export async function sendPing(host: string) {
-  return ping.promise.probe(host, { timeout: PING_TIMEOUT_S })
+  const cleanUrl: any = host.replace(/^https?:\/\//, '') // sanitize url
+  return ping.promise.probe(cleanUrl, { timeout: PING_TIMEOUT_S })
```
samsunoor, line 35: should timeout be configurable here?

either the default `response-time-greater-than-2-s` or from the config file

samsunoor (changes requested):

src/utils/ping.ts
sapiderman: hmm i dont think it should be a user configurable setting, don't see a need for it. It's an dev setting, set here so its not too long.

samsunoor (approved):

src/components/config/validate.ts
```diff
@@ -288,7 +290,13 @@ export const validateConfig = (configuration: Config): Validation => {
 
       if (!url) return PROBE_REQUEST_NO_URL
 
-      if (url && !isValidURL(url)) return PROBE_REQUEST_INVALID_URL
+      if (url) {
```
gsukanto, line 293: I think this `if (url)` is unnecessary since line 291 has covered this logic.

src/components/config/validate.ts
```diff
@@ -288,7 +290,13 @@ export const validateConfig = (configuration: Config): Validation => {
...
+        if (request.ping === true) {
```
gsukanto, line 294: how about having and instead of empty if like this?
for example: `if (request.ping !== true && !isValidURL(url))`

src/components/probe/index.ts
```diff
@@ -168,13 +168,14 @@ export async function doProbe(
...
-          0: 'URI not found',
-          1: 'Connection reset',
-          2: 'Connection refused',
-          3: 'Unknown error',
-          599: 'Request Timed out',
+          0: 'URI not found', // axios error
+          1: 'Connection reset', // axios error
+          2: 'Connection refused', // axios error
+          3: 'Unknown error', // axios error
+          4: 'ping timed out', // ping error
```
gsukanto, line 177: need to be Title case ex.: `Ping timed out`

gsukanto (approved):

nicnocquee (approved):

PR #620, opened 2022-02-25
chore: send recovery/incident after the first loop to Symon

kevin-hyperjump:
This PR solves #578


nicnocquee (approved):

dennypradipta (approved):

src/components/notification/process-server-status.ts
```diff
@@ -156,9 +158,12 @@ export const processThresholds = ({
       alertQuery: alert.query,
       state: state.value as 'UP' | 'DOWN',
       shouldSendNotification:
+        isFirstTimeSendEvent ||
```
haricnugraha, line 161: I think it will send a recovery notification if the first event is UP, could you check it again?

kevin-hyperjump: Nope, it will send even the first event DOWN

Screen Shot 2022-02-25 at 15 42 54

haricnugraha: https://user-images.githubusercontent.com/15191978/155685964-6610f540-fdb1-442f-b8a1-8a7a1e84baf4.mp4

src/components/notification/process-server-status.ts
```diff
@@ -28,6 +28,8 @@ import { Probe } from '../../interfaces/probe'
 import { ServerAlertState } from '../../interfaces/probe-status'
 import { ValidatedResponse } from '../../plugins/validate-response'
 
+let isFirstTimeSendEvent = true
```
nicnocquee, line 31: Just realized something. I think there's a bug in this implementation. I think it only sends the event for the first probe. Instead of adding isFirstTimeSendEvent variable here, maybe you should create another state machine to keep the isFirstTimeSendEvent state for all the probes.

nicnocquee (changes requested):

src/components/notification/process-server-status.ts
kevin-hyperjump: Solved!

nicnocquee (approved):

Issue #568, opened 2022-01-11
TCP server probing


nicnocquee:
There is a request to support probing a TCP server. I'm thinking to add this in the configuration

```yml
probes:
  - id: '1'
    name: Name of the probe
    description: Probe to check TCP server
    interval: 30
    timeout: 10
    socket:
      host: <ip>
      port: <port number>
      data: <string to send>
   alerts:
     - query: response.size > 0
       message: Response size is {{ response.size }}, expecting more than 0
 
```

PR #642, opened 2022-04-12
Fix TCP probing for symon

nicnocquee:
# Monika Pull Request (PR)  

1. #568 The issue has been closed but there's still problem connecting to Symon.

1.  Refactor many parts of the code, for example the threshold processing is now using the same function with the HTTP requests probe.
2. Save the result of probing TCP server to sqlite db.
3. For now, the alert for TCP probe is only when the server is reachable or not. When it's not reachable, the status code is set to 0, otherwise the status code is set to 200.

## How to test  
Stand alone mode:

1. Run a TCP node server. You can use this Node.js TCP server https://gist.github.com/nicnocquee/7ead55a4b2844803241b8cfed9da479f
2. Create a monika.yml filled with TCP probes just like in the example https://monika.hyperjump.tech/guides/probes#tcp
3. Run `npm run start -- -c ./monika.yml`

Symon Mode:

1. Run symon from the PR in our repo.
2. Create a TCP probe.


sapiderman (approved):
works for me!

image

haricnugraha (approved):

Issue #528, opened 2021-11-12
Monika crashes when config file (yml) is modified on Windows 11


sapiderman:
**Describe the bug**
Monika crashes when it detects a change in the YAML config file.

**To Reproduce**
1. Run monika using monika.yml `monika -c monika.yml`
2. While monika is running update the config file
3. Save the file
4. See monika crashes

**Expected behavior**
Monika should detect the change, reload the config file and keep on chugging

**Screenshots** 
See below

**Desktop (please complete the following information):**
 - Edition	Windows 11 Home Single Language
 - Version	21H2
 - OS build	22000.318

**Additional context**
- Running on windows native
- Running on Monika production
- Monika version `@hyperjumptech/monika/1.6.7 win32-x64 node-v16.13.0`

![image](https://user-images.githubusercontent.com/964106/141465719-e8b03c61-535a-471b-a589-d8f84cfdc22f.png)

sapiderman:
OOoh it happened the other day on Ubuntu too... seems there is a window where the config YAML is being updated just as it was being read by monika?

![Screenshot from 2022-02-16 01-36-02](https://user-images.githubusercontent.com/964106/154196931-abbc5039-527b-4561-92f6-9869e8f10dcc.png)

PR #653, opened 2022-04-19
Add readfile retries when re-reading configuration file

sapiderman:
# Monika Pull Request (PR)  

1. Improves on issue #528 
2. When the config file is updated, monika reloads the config
3. But if the file is still being updated, monika may fail

1.  Add a retry loop when reading configfile
2. Add a delay between reads (about 700ms)
3. 700ms delay was chosen after several tweaks and tests

## How to test  

Method 1 : automated
1. Run monika with the config file "testchange.yml" ex: `monika -c testchange.yml -s 0`
4. Then (in another shell) run the config changer loop script : `npx ts-node ./test/others/change-config-loop.ts`
5. You should get an exception "Failed to read testchange.yml. Got empty"

Before the fix: Using change-config-loop script to modify config file:
image

Method 2: Manual
1. Run monika with any config file you want
2. Using an editor, while monika is running, change the config file. And Save.

After the fix, after a few thousand iterations, still not glitching:
image
 
## Some Testing Notes
1. This seems to be very environment dependent. On my daily machine (Pop_OS 21.10) it rarely happened, even with the loop changer script.
2. On a Windows machine it happened quite easily. I had the sources on WSL, but editing from vscode on the main Windows environment.
3. Updating the config file on vscode seems to take forever on Windows, which results int he 10x loop retry implementation.

Debugging on reload retries. Alternative len a debug output, shows file length and reread attempts.
image


src/components/config/parse.ts
```diff
@@ -32,17 +32,40 @@ import parseInsomnia from './parse-insomnia'
...
+      let tries = 10 // tries multiple times to load the file
+      while (configString.length === 0 && tries > 0) {
+        sleep(700)
+        configString = readFileSync(source, 'utf-8')
+        if (configString.length > 0) {
+          break
+        }
+        tries--
+      }
+      // await sleep(300) // wait a bit and re-read the config file again
```
nicnocquee, line 65: Remove this?

nicnocquee (changes requested):

PR #669, opened 2022-05-23
Create deployment pipeline for Homebrew

raosan:
# Monika Pull Request (PR)  

1. Add additional github action to publish package to [Homebrew Core Repository](https://github.com/Homebrew/homebrew-core) so monika can be installed via Homebrew
2. Fix typos
3. Add example for monika.rb file

1. Adding publish-homebrew.yml

## How to test  
1. In Actions menu, there will be "Publish to Homebrew" action. it can be run on release or manually
2. After it run, check the [Homebrew Core Repository PR list](https://github.com/Homebrew/homebrew-core/pulls), there will be a new PR from monika


.github/workflows/publish-homebrew.yml
```diff
@@ -0,0 +1,22 @@
+name: Publish to Homebrew
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
```
nicnocquee, line 6: Since the download url is from npm registry, when a release is created, the `mislav/bump-homebrew-formula-action` action might be executed **before** the other workflow finishes uploading to npm. Thus, this workflow will fail.

In this case, instead of release trigger, I think it's better to use [workflow_run trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run).

nicnocquee (changes requested):

.github/workflows/publish-homebrew.yml
raosan: updated

nicnocquee (approved):

Issue #679, opened 2022-05-30
Fix warnings


nicnocquee:
GitHub checks still shows many warnings. Let's fix it.

PR #694, opened 2022-06-09
DRAFT:Bugfix-679/clean warning message

lukman7788:
# Monika Pull Request (PR)  

1. cleaning warning message at monika main project (issue [#679](https://github.com/hyperjumptech/monika/issues/679))
2. fixing error build next.js at docs directory


nicnocquee (changes requested):
Few comments:

1. Instead of disabling the rule in the next line, maybe better to fix the code by following the recommendation. For example, [for the no-for-loop warnings, change the code to use for-of](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-for-loop.md).
2. If changing the code is not possible or too much changes, better to turn off the rule in .eslintrc.
3. Can you split the PR into three: for fixing the files in `docs`, `src`, and `test`.

Issue #710, opened 2022-06-16
Ignore docs from code coverage


nicnocquee:
Since the docs is not really part of the application, we should ignore the directory from the codecov. For more information, refer to the [Ignoring a specific folder](https://docs.codecov.com/docs/ignoring-paths#ignoring-a-specific-folder) documentation in Codecov.

nicnocquee:
it is already ignored.

PR #720, opened 2022-06-21
Ignore Docs in Codecov

sapiderman:
# Monika Pull Request (PR)  

1. Solves #710

## After PR  
1.  Ignores the /docs path for codecov scanner


sapiderman:
It looks like Codecov is only covering /src

PR #784, opened 2022-07-20
fix: TCP request doesn't make any request

### LINKED ISSUE #783: TCP request doesn't make any request

kevin-hyperjump:
This PR fixes #783

Simply, we should immediately end the connection after successfully connecting to the server to avoid a timeout error, because the `"data"` will not listen if `client.write` is not emitted.


nicnocquee (approved):

sapiderman (changes requested):
Can we print out the TCP errors out to the console (through pino) like regular http errors when we get them. Otherwise how would users know they're getting errors?

kevin-hyperjump:
> Can we print out the TCP errors out to the console (through pino) like regular http errors when we get them. Otherwise how would users know they're getting errors?

The TCP error was handled by Promise rejection, I guess it will be printed too. If not, please create a new issue, I'll keep this PR only for fixing the bug :)

sapiderman (approved):

sapiderman:
> > Can we print out the TCP errors out to the console (through pino) like regular http errors when we get them. Otherwise how would users know they're getting errors?
> 
> The TCP error was handled by Promise rejection, I guess it will be printed too. If not, please create a new issue, I'll keep this PR only for fixing the bug :)

okay!

Issue #809, opened 2022-07-22
Add validation for timeout value


nicnocquee:
There are several addition checks we need to make sure the probing process is correct:

~- If `interval` is set, but the requests' timeout is not set, the `interval` value **should be more than 10 seconds multiplied by the number of requests in the probe** because the default value for the requests' timeout is 10000 milliseconds. If the condition is not satisfied, show the following error: `The interval in the probe with name  "${probeName} should be greater than ${10 * numOfRequests} because the default request's timeout is 10 seconds. Current interval value is ${interval} seconds but the expected value is greater than ${10 * numOfRequests} seconds.`~
- [x] If `interval` is set, and the requests' timeout is set, the `interval` value **should be greater than the total `timeout` values**. Note that the requests in a probe can have different timeout values. If the condition is not satisfied, show the following error: `The interval in the probe with name  "${probeName}" should be greater than the total timeout value of all requests in this probe (${totalTimeout} seconds). Current interval value is ${interval} seconds but the expected value is greater than ${totalTimeout} seconds.`
- [x] If `interval` is not set, and the requests' timeout is set, the requests' timeout value **should be less than 10000** because the default value for `interval` is 10 seconds. If the condition is not satisfied, show the following error: `The total timeout values of all requests in the probe with name "${ probeName}" should be less than 10 seconds. Current interval value is ${interval} seconds (default value) but the timeout values of this probe is ${totalTimeout} seconds. To fix this, define the interval value to be greater than ${totalTimeout} seconds or reduce the timeout values of each requests so that the sum of all timeout is less than 10 seconds.`

PR #832, opened 2022-07-27
feat: add validation for timeout value

kevin-hyperjump:
This PR solves #809 

- If the `interval` value and requests' timeout are not set, the `interval` value will be set to `requests.length * 10` and timeout will be set to `10000ms` = `10s`

```yml
probes:
  - id: '1'
    name: GitHub
    requests:
      - url: https://github.com
      - url: https://github.com/hyperjump
```

- If the `interval` value is set and requests' timeout are not set, the `interval` value will not change and timeout will be set to `10000ms` = `10s`

```yml
probes:
  - id: '1'
    name: GitHub
    interval: 10
    requests:
      - url: https://github.com
      - url: https://github.com/hyperjump
```

The error message if the `totalTimeoutSeconds > interval`:

Screen Shot 2022-07-27 at 11 49 13


src/components/config/validate.ts
```diff
@@ -320,6 +320,17 @@ export const validateConfig = (configuration: Config): Validation => {
 
     if ((!socket && (requests?.length ?? 0)) === 0) return PROBE_NO_REQUESTS
 
+    // Validate Interval
+    const timeouts = requests.map((req) => req.timeout)
+    const totalTimeout = timeouts.reduce((a, b) => a + b)
```
nicnocquee, line 325: could be shorter and 1 less loop:

```typescript
const totalTimeout = requests.reduce((prev, curr) => prev.timeout + curr.timeout, 0)
```

test/index.test.ts
```diff
@@ -52,8 +52,10 @@ describe('monika', () => {
     .do(() =>
       cmd.run(['--config', resolve('./test/testConfigs/noInterval.yml')])
     )
-    .it('runs with config without interval', (ctx) => {
-      expect(ctx.stdout).to.contain('Starting Monika.')
+    .catch((error) => {
+      expect(error.message).to.contain(
+        'The total timeout values of all requests in the probe with name "Example" should be less than 10 seconds. Current interval value is 10 seconds (default value) but the timeout values of this probe is 10000 seconds. To fix this, define the interval value to be greater than 10 seconds or reduce the timeout values of each requests so that the sum of all timeout is less than 10 seconds.'
```
nicnocquee, line 57: I don't understand how this test can ever pass. It has the words `(default value)`, but those words are nowhere to be found in the main code. Not even in the src/components/config/validate.ts file. 

What did I miss?

nicnocquee (changes requested):

src/components/config/validate.ts
```diff
@@ -304,12 +304,12 @@ export const validateConfig = (configuration: Config): Validation => {
...
 
   // Validate probes
-  if ((configuration?.probes?.length ?? 0) === 0) return NO_PROBES
+  if (probes.length === 0) return NO_PROBES
 
   // Check probes properties
   for (const probe of probes) {
-    const { alerts = [], requests, socket } = probe
-    const socketAlerts = probe.socket?.alerts ?? []
+    const { name, interval, alerts = [], requests, socket } = probe
```
nicnocquee, line 311: I think we also need to validate the interval to be greater than 0

src/components/config/validate.ts
```diff
@@ -320,6 +320,17 @@ export const validateConfig = (configuration: Config): Validation => {
 
     if ((!socket && (requests?.length ?? 0)) === 0) return PROBE_NO_REQUESTS
 
+    // Validate Interval
+    const timeouts = requests.map((req) => req.timeout)
+    const totalTimeout = timeouts.reduce((a, b) => a + b)
+    const totalTimeoutSeconds = totalTimeout / 1000
+
+    if (totalTimeoutSeconds > interval) {
```
nicnocquee, line 328: Also need to make sure each timeout is greater than 0

nicnocquee (changes requested):

test/index.test.ts
kevin-hyperjump: Updated!

src/components/config/validate.ts
kevin-hyperjump: Updated!

src/components/config/validate.ts
kevin-hyperjump: Updated!

src/components/config/validate.ts
kevin-hyperjump: Updated!

test/index.test.ts
kevin-hyperjump: Sorry that's an invalid test since the Monika doesn't throw any error the test runner doesn't catch anything and skips the test. I will remove the test :)

nicnocquee (approved):

nicnocquee (approved):

PR #856, opened 2022-08-12
Make STUN non blocking

### LINKED ISSUE #848: When STUN is unavaiable, probes fail to run
**Describe the bug**
Probe loop fail to run when STUN server is inaccesible

**To Reproduce**
1. Run Monika on docker
2. Have one or more probes in monika.yaml
3. Block STUN

**Expected behavior**
1. STUN unaccesible should not be fatal
2. Probe loops should keep working

**Screenshots**

image

**Desktop (please complete the following information):**
 - OS: MacOS
 - Monika Version: Latest docker image

**Additional context**
Reference: From the discussions [here](https://github.com/hyperjumptech/monika/discussions/847)

sapiderman:
# Monika Pull Request (PR)  

1. Closes #848 
2. Improve how flags are passed into functions

1. Refactor function parameters using objects
2. Add monika flags as paramenter inputs for easier checking/access
3. Checks for -s == -1 to disable STUN blocking

## How to test  
1. First we want to block stun server. Find out the stun server available from your provider/location. Ping the stun to get its IP like below (your ping and ip result will likely be different than the example).
```bash
ping stun.l.google.com

PING stun.l.google.com (172.253.118.127) 56(84) bytes of data.
64 bytes from sl-in-f127.1e100.net (172.253.118.127): icmp_seq=1 ttl=106 time=22.5 ms
64 bytes from sl-in-f127.1e100.net (172.253.118.127): icmp_seq=2 ttl=106 time=22.8 ms
64 bytes from sl-in-f127.1e100.net (172.253.118.127): icmp_seq=3 ttl=106 time=23.6 ms
64 bytes from sl-in-f127.1e100.net (172.253.118.127): icmp_seq=4 ttl=106 time=26.9 ms
```
2.  Next block this ip. In my case I'm using Linux UFW firewall, to block a specific ip address, type below:
```bash
sudo ufw deny out from any to 172.253.118.127
```
4. Now run Monika again with the `-s -1` parameter.
5. Expect that STUN is still inaccessible, but the probing will continue to work. See 3rd screenshot below.

## Screenshot

Monika with `-s 5`, checking stun every 5 seconds.
image

Monika with `-s 5` but with STUN blocked. Notice the probe not running.
image

Monika with `-s -1` to unblock STUN checking. Notice the `stun inaccessible` message, but probe continues.
image

### DIFF
--- a/docs/src/pages/guides/cli-options.md
+++ b/docs/src/pages/guides/cli-options.md
@@ -273,9 +273,9 @@ monika --follow-redirects 0 # disable following redirects
 
 ## STUN
 
-By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server every 20 second intervals. Continuously STUN checking ensures that connectivity to the outside world is guaranteed. When the network is down, STUN checking fails, probing will be paused.
+By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server every 20 second intervals. Continuously STUN checking ensures that connectivity to the outside world is guaranteed. When STUN checking fails, Monika assumes the network is down and probing will be paused.
 
-You can specify the number of intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the parameter below:
+You can specify the number of checking intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to every 10 seconds type the parameter below:
 
 ```bash
 monika -s 10
@@ -283,7 +283,7 @@ monika -s 10
 
 If the number is zero or less, monika will check the STUN server just once, not repeatedly, to get public IP.
 
-For internal networks where no outside connection is provided, you can disable the STUN checking. To disable STUN checking, set -s to `-1`.
+For internal networks where no outside connection is needed, you can disable the STUN checking by setting the `-s` flag to `-1` as follows.
 
 ```bash
 monika -s -1
--- a/docs/src/pages/guides/cli-options.md
+++ b/docs/src/pages/guides/cli-options.md
@@ -273,13 +273,21 @@ monika --follow-redirects 0 # disable following redirects
 
 ## STUN
 
-By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server in 20 second intervals. You can specify the number of intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the command below:
+By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server every 20 second intervals. Continuously STUN checking ensures that connectivity to the outside world is guaranteed. When the network is down, STUN checking fails, probing is paused.
+
+You can specify the number of intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the parameter below:
 
 ```bash
 monika -s 10
 ```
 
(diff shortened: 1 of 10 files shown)


docs/src/pages/guides/cli-options.md
```diff
@@ -273,14 +273,22 @@ monika --follow-redirects 0 # disable following redirects
 
 ## STUN
 
-By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server in 20 second intervals. You can specify the number of intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the command below:
+By default monika will continuously check the [STUN](https://en.wikipedia.org/wiki/STUN) server every 20 second intervals. Continuously STUN checking ensures that connectivity to the outside world is guaranteed. When the network is down, STUN checking fails, probing will be paused.
```
nicnocquee, line 276: > When the network is down, STUN checking fails, probing will be paused.

When STUN checking fails, probing will be paused because Monika assumes that the network is done.

docs/src/pages/guides/cli-options.md
```diff
@@ -273,14 +273,22 @@ monika --follow-redirects 0 # disable following redirects
...
+
+You can specify the number of intervals using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the parameter below:
```
nicnocquee, line 278: You can specify the checking interval using the `-s` or `--stun` flags followed by a number in seconds. For example to set the interval to 10 seconds type the command below:

docs/src/pages/guides/cli-options.md
```diff
@@ -273,14 +273,22 @@ monika --follow-redirects 0 # disable following redirects
...
 
 ```bash
 monika -s 10
 ```
 
 If the number is zero or less, monika will check the STUN server just once, not repeatedly, to get public IP.
 
+For internal networks where no outside connection is provided, you can disable the STUN checking. To disable STUN checking, set -s to `-1`.
```
nicnocquee, line 286: For internal networks where no outside connection is needed, you can disable the STUN checking by setting the `-s` flag to `-1` as follows.

src/looper/index.ts
```diff
@@ -122,6 +124,9 @@ export function isIDValid(config: Config, ids: string): boolean {
 }
 
 export async function loopCheckSTUNServer(interval: number): Promise<any> {
+  // if stun is disabled, no need to create interval
+  if (interval === -1) return
```
nicnocquee, line 128: isn't this the only change needed to make the `-s -1` flag works? The other changes should be in different PR.

nicnocquee (changes requested):

src/looper/index.ts
sapiderman: another change required in loopProbe() related to looping check when STUN is connected:
now its:    (isConnectedToSTUNServer && !isPaused) ||  flags.stun === DISABLE_STUN

package.json
```diff
@@ -184,5 +184,8 @@
       "node14-win-x64"
     ],
     "outputPath": "dist"
+  },
+  "volta": {
+    "node": "16.14.2"
```
samsunoor, line 189: - Is this intended ? Monika contribution guide does not explain Volta as its standard environment
- Node v16 does not reflect `pkg`'s targets

sapiderman: yes, node v16.xx is the recommended target build env.

nicnocquee (approved):

PR #866, opened 2022-08-25
fixed summary bug

zgolus:
# Monika Pull Request (PR)  

It fixes a bug which occurs when running `monika --summary`:
```
> monika --summary
ERROR: Summary notification: Cannot read property 'probes' of undefined
```

I only added a check if `config` is defined in `getSummary()` function. It seems to be undefined when running the function in CLI. The value on number of probes in CLI is taken for PID file, so it's still OK if config is undefined.

## How to test  
Just run `monika --summary`

### DIFF
--- a/src/components/logger/history.ts
+++ b/src/components/logger/history.ts
@@ -477,7 +477,7 @@ export async function getSummary(): Promise<any> {
   const config = getConfig()
 
   return {
-    numberOfProbes: config.probes.length,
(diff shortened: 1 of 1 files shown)


sapiderman (approved):
Good PR! :beer: :clinking_glasses: :champagne:

worked for me (before/after patch)
image

nicnocquee (approved):

PR #873, opened 2022-09-02
Monika doesn't save requests to local db and can't send report to Symon

### LINKED ISSUE #871: Monika doesn't save requests to local db and can't send report to Symon
`getContext` doesn't work if initialized in global, it must be initialized inside the `doProbe` function.

Screen Shot 2022-09-02 at 14 00 07

~And the flag `verboseLogs` will always return `undefined` since the available verbose's flags are either `verbose` or `keep-verbose-logs`.~

kevin-hyperjump:
This PR fixes #871

### DIFF
--- a/src/looper/index.ts
+++ b/src/looper/index.ts
@@ -154,10 +154,7 @@ type loopProbeParams = {
  */
 function loopProbe({ probe, notifications }: loopProbeParams) {
   let counter = 0
-  const flags = getContext().flags
-  // flags is the monika parameter flags... these are used:
-  // flags.repeats: number,
-  // flags.followRedirects: number
+  const { flags } = getContext()
 
   const probeInterval = setInterval(() => {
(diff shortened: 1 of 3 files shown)


src/looper/index.ts
```diff
@@ -157,7 +157,6 @@ function loopProbe({ probe, notifications }: loopProbeParams) {
   const flags = getContext().flags
   // flags is the monika parameter flags... these are used:
```
sapiderman, line 158: I think we dont need these comments anymore. we can cleanup and erase this?

sapiderman (approved):
LGTM 
:coffee: :tea: :milk_glass:

src/looper/index.ts
kevin-hyperjump: Sure, I'll remove them.

nicnocquee (approved):

Issue #859, opened 2022-08-12
Monitor Postgres


nicnocquee:
As mentioned in [this discussion](https://github.com/hyperjumptech/monika/discussions/843), it'd be great if Monika can also check if a postgres is running. Maybe using [this function](https://node-postgres.com/api/client#clientconnect):

```javascript
const { Client } = require('pg')
const client = new Client({
  host: 'my.database-server.com',
  port: 5334,
  user: 'database-user',
  password: 'secretpassword!!',
})
client.connect(err => {
  if (err) {
    console.error('connection error', err.stack)
  } else {
    console.log('connected')
  }
})
```

```yaml
probes:
  - id: 'check-redis'
    postgres:
      host: '<hostname>'
      port: <port>
      password: '<password>'
      user: <user>
```

Just like `tcp` and `ping` probing that we already have, user doesn't need to define any alerts for the `postgres` checking.

So, todo:

- [x] Monitor Postgres, send notification when it cannot be reached after reaching threshold.
- [x] Update [monika-config-schema.json](https://github.com/hyperjumptech/monika/blob/main/monika-config-schema.json) to include the new properties.
- [x] Add in the documentation.

sapiderman:
re-opening as the main function still need to PR

PR #892, opened 2022-09-26
Add postgres probe type

sapiderman:
# Monika Pull Request (PR)  

1. Implements #859
2. Add postgres as a probe type

1.  Using pg 'node-postgres' lib
2. Add new interface and postgres as probe type

## How to test  
1. Use the following test config `test.yaml`:
```yaml
notifications:
  - id: unique-id
    type: desktop
probes:
  - id: '1'
    name: Example
    description: Probe
    interval: 5
    postgres:
      # - host: 172.16.1.10
      #   port: 5432
      #   database: mydb
      #   username: dbuser
      #   password: dbpassword
      - uri: 'postgres://uriuser:uripassword@172.77.1.1:5432/uri-db'

    incidentThreshold: 2
    recoveryThreshold: 2
    alerts:
      - assertion: response.status != 200
        message: 'AAALEEEERT!!!'

```
2. Spin up a local postgres instance with the following example docker yaml: `docker compose up -d`
```yaml
  my-postgres:
    image: "postgres:latest"
    container_name: my-postgress
    ports:
      - 5432:5432
    environment:
      POSTGRES_PASSWORD: db-password
      POSTGRES_USER: db-user
      POSTGRES_DB: my-database
    volumes:
      - ./db/data:/var/lib/postgresql/data
    restart: unless-stopped
``` 
3. Run monika to probe the db: `bin/run -c test.yaml -s -1`
4. Expect logging to run with `postgres ok` or any errors if connection interrupted
5. You can provide either a uri connection string or a host/port/database type config fields.

``` yaml
postgres:
  - database: mydb
    host: 1.1.1.1
    uri: 'postgres://uriuser:uripassword@172.77.1.1:5432/uri-db''
```
This will work even though port is not provided (and is required) the presence of a uri satisfies connection config.

## Screenshot
Monika test logs: started with the db off, then started, then turned off again.
image

Case when postgres is unavailable, params from host,port,username, password fields vs params from uri (bottom)
image


docs/src/pages/guides/probes.md
```diff
@@ -89,6 +89,24 @@ probes:
...
+```yaml
+probes:
+  - id: 'postgres-01'
+    name: database health
+    description: ensure db health
+    interval: 30 # in seconds
+    postgres:
+      - host: 172.15.0.1
+        port: 5432
+        database: my-database
```
dennypradipta, line 105: Was there any possibility that I could probe my database without having to supply a database name? Just to check the connection, for example

sapiderman: It is possible to have no database with this driver.. but the user / password is expected to be string. In the other app we were looking at, database IS mandatory, but defaults to a 'template1' if not specified.

lukman7788 (approved):

dennypradipta (approved):

nicnocquee:
This doesn't seem to work. I used this config

```yaml
probes:
  - id: 'postgres-01'
    name: database health
    description: ensure db health
    interval: 10 # in seconds
    postgres:
      - host: 172.15.0.1
        port: 5432
        database: my-database
        password: db-password
```

And no postgres running. Then this happened after a long time:

Screenshot 2022-10-05 at 08 53 58

Monika crashed.

nicnocquee (changes requested):
Doesn't seem to work.

sapiderman:
> Doesn't seem to work.

fixed. Was releasing a client `client.release()` that was never connected, ooops.

docs/src/pages/guides/probes.md
```diff
@@ -89,6 +89,24 @@ probes:
         ping: true
 ```
 
+### Postgres Request
+
+Monika provides a way to check your postgres database's health with a 'postgres' type probe. You can define the `host`, `port` and `user` and `password` like this:
+
+```yaml
+probes:
```
haricnugraha, line 97: The common way to connect to a Postgres database is by using connection URI. It would be better if we can accept it as well. More info: https://www.prisma.io/dataguide/postgresql/short-guides/connection-uris

dennypradipta: +1 for this. Maybe if the `uri` is not empty, the user won't have to fill out `host`, `port`, `user,` and `password`. Vice versa

sapiderman: Strangely enough `URI` connection string is not supported natively by the node-postgres library. So the URI string is now parsed into `host`, `port` etc.. and passed to the constructor.

nicnocquee (approved):

PR #918, opened 2022-10-13
Feat: MongoDB Probe Request

### LINKED ISSUE #914: Monitor MongoDB Client
It will be great if we can monitor MongoDB clients.

We can use `db.stats().ok;` or `db.runCommand({ ping: 1 });` query to check if the `ok` value is equal to 1
image

The input would be something like this:

```
  - id: 'check-mongodb'
    mongodb:
      url: <mongodb_connection_url> # required if the host, port, password, and user is empty
      host: '<host>' # required if url is empty
      port: <port> # required if url is empty
      password: '<password>' # required if url is empty
      user: <user> # required if url is empty
```

You may refer to [Redis Client probing.](https://github.com/hyperjumptech/monika/blob/main/src/components/redis-request/index.ts)

dennypradipta:
# Monika Pull Request (PR)  

This PR closes #914 

Implemented a MongoDB request based on Redis request

## How to test  
1. Spin up a Docker instance by running these:
```bash
# Mongo without auth
docker run -d --name mongo-test -p 27017:27017 mongo:latest

# Mongo with auth
docker run -d --name mongo-test -p 28017:27017 -e MONGO_INITDB_ROOT_USERNAME=mongoadmin -e MONGO_INITDB_ROOT_PASSWORD=secret mongo:latest
```

2. Use this config:
```yaml
notifications:
  - id: desktop
    type: desktop
probes:
  - id: '1'
    name: URI without credential
    mongo:
      - uri: mongodb://localhost:27017
  - id: '2'
    name: Manual without credential
    mongo:
      - host: localhost
        port: 27017
  - id: '3'
    name: URI with credential
    mongo:
      - uri: mongodb://mongoadmin:secret@localhost:28017
  - id: '4'
    name: Manual with credential
    mongo:
      - host: localhost
        port: 28017
        username: mongoadmin
        password: secret
```
Video available at:
https://streamable.com/p4m2qu


src/monika-config-schema.json
```diff
@@ -83,6 +83,51 @@
...
+            "type": "string",
+            "examples": ["localhost", "127.0.0.1"],
+            "default": ""
+          },
+          "port": {
+            "title": "Port number",
+            "description": "Port number used for the MongoDB",
+            "type": "integer",
+            "examples": ["27017"],
+            "default": ""
```
sapiderman, line 114: would it be good to have default to 27017?

sapiderman (approved):
Nicely done! All works
:bowtie: :feelsgood: :godmode:


image

src/monika-config-schema.json
dennypradipta: Good idea. Updated

kevin-hyperjump (approved):

PR #929, opened 2022-10-21
fix: monika parses integer to string

### LINKED ISSUE #922: JSON.stringify() converts numeric value into string while passing value from body
body:
          otp_id: '3100791963'
          otp: 359496
     
It is coverting otp as a string. But in my case it has to be numeric value only.
image

haricnugraha:
# Monika Pull Request (PR)  

1. Fix Monika parses integer in the request body to a string. This will resolves #922.

1.  Create a function that generates chaining request body
2. Test the function

## How to test  
1. Create a http server that receives request with body that consists of property with integer data type
2. Check if Monika send the request with the correct data type

### Monika Config Sample
Screenshot from 2022-10-21 18-31-34

### Before
Screenshot from 2022-10-21 18-31-34

### After
Screenshot from 2022-10-21 18-32-39

### DIFF
--- a/src/components/http-request/http-request.test.ts
+++ b/src/components/http-request/http-request.test.ts
@@ -28,9 +28,12 @@ import { setupServer } from 'msw/node'
 import { RequestInterceptor } from 'node-request-interceptor'
 import withDefaultInterceptors from 'node-request-interceptor/lib/presets/default'
 import { setContext } from '../../context'
-import { RequestConfig } from '../../interfaces/request'
+import type {
+  ProbeRequestResponse,
+  RequestConfig,
+} from '../../interfaces/request'
 
-import { httpRequest } from '.'
+import { generateRequestChainingBody, httpRequest } from '.'
 
 describe('probingHTTP', () => {
   let interceptor: any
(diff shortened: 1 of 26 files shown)


src/components/http-request/http-request.test.ts
```diff
@@ -352,4 +355,91 @@ describe('probingHTTP', () => {
...
+      it('should generate request chaining body', () => {
+        // arrange
+        type TestTable = {
+          body: Record<string, any>
+          responses: ProbeRequestResponse[]
+          expected: Record<string, any>
+        }
+        const testTables: TestTable[] = [
+          {
+            body: {
```
nicnocquee, line 370: Can you please also add a test for when the body is just string, not an object.

nicnocquee (changes requested):

src/components/http-request/http-request.test.ts
haricnugraha: Ok. Working on it.

haricnugraha: Added here [60c1bb7](https://github.com/hyperjumptech/monika/pull/929/commits/a9696b888163ef69bf635e409dce067c85af631d).

nicnocquee (approved):

dennypradipta (approved):

PR #950, opened 2022-12-02
chore: test with Node.js v18 in CI

ariya:
Also, cache npm dependencies.


.github/workflows/node.js.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
         strategy:
             matrix:
-                node-version: [14.x, 16.14]
+                node-version: [14.x, 16.x, 18.x]
```
ariya, line 19: Also, fixes `16.14` which doesn't make sense.

sapiderman: hmm could it be due to this Fetch API?

```bash
(node:113293) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
```

sapiderman: It seems the msw is not properly supporting node v.18 yet. 
As per this open issue: https://github.com/mswjs/msw/issues/1388

sapiderman: This issue is fixed in #967... closing this PR

PR #952, opened 2022-12-05
[Symon] Monika always send recovery at first time run even if the sit…

### LINKED ISSUE #951: [Symon] Monika always send recovery at first time run even if the site is incident
**To Reproduce**

1. Run Monika with an incident site (Makes sure it's on Symon mode)
2. An incident will appear on the event table
3. Stop Monika and run again
4. Incident changed to recovery (even if the site is still incident)

kevin-hyperjump:
Fixes #951

### DIFF
--- a/src/components/probe/index.ts
+++ b/src/components/probe/index.ts
@@ -23,9 +23,13 @@
  **********************************************************************************/
 
 import * as mongodbURI from 'mongodb-uri'
+import { parse } from 'pg-connection-string'
+
+import { getContext } from '../../context'
 import events from '../../events'
 import { Notification } from '../../interfaces/notification'
 import { Probe } from '../../interfaces/probe'
+import { ServerAlertState } from '../../interfaces/probe-status'
 import { ProbeRequestResponse } from '../../interfaces/request'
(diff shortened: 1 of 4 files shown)


src/components/probe/index.ts
```diff
@@ -144,14 +138,13 @@ function checkThresholdsAndSendAlert(
 }
 
 export function getProbeStatesWithValidAlert(
-  probeStates: ServerAlertState[],
-  isSymonMode: boolean
+  probeStates: ServerAlertState[]
 ): ServerAlertState[] {
   return probeStates.filter(
     ({ isFirstTime, shouldSendNotification, state }) => {
       // ignore first up event for non Symon mode
```
nicnocquee, line 145: remove this comment, i don't think it's accurate anymore, right?

nicnocquee (changes requested):

src/components/probe/index.ts
kevin-hyperjump: Comment removed.

nicnocquee (approved):

sapiderman (approved):

PR #953, opened 2022-12-07
CI: Enable Some Linter Rules

haricnugraha:
# Monika Pull Request (PR)  

1. As a part of the effort to solve issue #679.

1. Enable the following rules and fix the violated codes
 
- no-warning-comments
- no-process-exit
- @typescript-eslint/no-use-before-define
- unicorn/no-lonely-if
- padding-line-between-statements
- unicorn/prefer-ternary
- dot-notation
- unicorn/numeric-separators-style
- unicorn/prefer-optional-catch-binding
- unicorn/consistent-destructuring
- unicorn/prefer-includes
- node/no-extraneous-import
- unicorn/prefer-spread
- unicorn/no-for-loop
- unicorn/import-style
- @typescript-eslint/ban-types
- unicorn/no-useless-undefined
- no-unsafe-optional-chaining
- unicorn/prefer-array-index-of
- unicorn/prefer-number-properties
- unicorn/catch-error-name
- unicorn/prefer-date-now
- default-param-last
- unicorn/prefer-set-has
- unicorn/no-process-exit
- unicorn/prefer-switch

## How to test  
1. Run `npm run posttest`


src/index.ts
```diff
@@ -632,6 +632,7 @@ process.on('SIGINT', async () => {
 
   em.emit(events.application.terminated)
 
+  // eslint-disable-next-line no-process-exit
```
sapiderman, line 635: While we're fixing code, is it better to throw here instead rather than disable the warning?

reference: https://eslint.org/docs/latest/rules/no-process-exit

sapiderman (approved):
looks better. 🍷 🍺 🍸

src/index.ts
haricnugraha: If the code is intended to stop the process and exit, disable the rule is a better option.

nicnocquee (approved):

PR #997, opened 2023-02-20
chore: improves `package.json` keywords

kevinhermawan:
Although seemingly small, even minor changes can have a significant impact, much like the [butterfly effect](https://en.wikipedia.org/wiki/Butterfly_effect) (although not in the context of chaos theory)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "Slow site notifications",
     "Resource monitoring",
     "Availability monitoring",
-    "Web health tracking"
+    "Web health tracking",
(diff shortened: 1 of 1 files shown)


dennypradipta (approved):

dennypradipta:
Approved, but maybe you should add "Synthetic monitoring" too

kevinhermawan:
> Approved, but maybe you should add "Synthetic monitoring" too

Ah, yes! We can add even more power to the wings.

sapiderman (approved):
👍🏽

Issue #999, opened 2023-02-23
Update "New notification" documentation


nicnocquee:
In the ["New Notification](https://monika.hyperjump.tech/guides/new-notifications)" page, I think we should also mention about updating the JSON schema in the validation section.

PR #1003, opened 2023-02-28
Add mention of json schema in documentations

sapiderman:
# Monika Pull Request (PR)  

1. Implements #999 

1.  Add mention of json schema when adding new notification
2. Gave schema example


ariawan (changes requested):
In addition to the in-code-validations above, Monika users have the option to use JSON schema validation from their favorite editors. This is super convenient and provides useful real-time feedback. Therefore it is mandatory to add a new schema to reflect your changes, otherwise, your new notification type will not be recognized and flagged as an unknown type.

nicnocquee (approved):

PR #1028, opened 2023-04-12
Chore: extract notification to workspace package

raosan:
# Monika Pull Request (PR)  

To make the code more modular, we need to extract notifications related code to separerate npm package so that it can be reused

1.  Extract the notification channels code to monorepo workspace in folder `packages/notification`
2. update dependencies
3. update import path

## How to test  
1. run `npm install`
2. run `npm run build -w packages/notification`
3. run monika `npm start` with defined config
4. see notification is sent as is

## Notes
Many changes is just updating the import path. don't be overwhelmed :)


packages/notification/channel/index.ts
```diff
@@ -63,10 +87,12 @@ export type NotificationMessage = {
     | NotificationStatusUpdateMessageMeta
 }
 
-export type NotificationChannel<T = any> = {
+// TODO: convert contentOrMessage to content, and implement getContent in every channel
+type NotificationChannel<T = any> = {
   validator: AnySchema
-  send: (notificationData: T, message: NotificationMessage) => Promise<void>
+  send: (notificationData: T, contentOrMessage: any) => Promise<void>
```
haricnugraha, line 93: Could you use `contentOrMessage` parameter has a stricter type? Because when we need to implement a new notification channel it will be hard if we don't know what type it is. And you need to update the documentation [here](https://monika.hyperjump.tech/guides/new-notifications#add-a-new-notification) if it change.

nicnocquee (approved):

packages/notification/channel/index.ts
raosan: updated by creating seperate function that accept content as parameter

haricnugraha (approved):

PR #1038, opened 2023-05-03
refactor: content footer

herenacreative:
# Monika Pull Request (PR)  

1. refactor content footer
2. Screen Shot 2023-05-03 at 10 16 45

## How to test  
1. cd docs/
3. npm run dev


docs/src/pages/index.js
```diff
@@ -627,28 +627,53 @@ export default function IndexPage() {
           />
           <div className="container20">
```
dennypradipta, line 628: image

The height is uneven because on the right card, it has 3 lines of title while the left has only 1. 

My suggestion is that this `container20` class should use CSS Grid, so that the height will be even. Refer to the guide here: https://css-tricks.com/snippets/css/complete-guide-grid/

dennypradipta: Also, might need some `padding-bottom` because it is placed so close to the footer.

dennypradipta (changes requested):

docs/src/pages/index.js
herenacreative: okay, updated

docs/src/pages/index.js
```diff
@@ -627,28 +627,53 @@ export default function IndexPage() {
...
-                    event({
-                      action: 'cta_button_footer',
-                      category: 'cta_button',
-                    })
-                  }
-                >
-                  Start Now
-                </a>
-              </Link>
+              <h2 className="text-white md:text-left lg:text-32 text-2xl text-center font-bold">
```
dennypradipta, line 654: image

Another improvement: The text should be at the top, while the button should be at the bottom.

Example:
image

So no matter how long the text is, the button is always at the bottom.

dennypradipta (changes requested):

dennypradipta (approved):

sapiderman (approved):
👍🏽 👍🏽  looks ok. approved when you get all the conversations resolved.

PR #1061, opened 2023-06-14
Feat: Add simple success/fail result to each probe and new prometheus metric

jgillick:
# Monika Pull Request (PR)  

1. Adds a simple boolean success/fail indicator to the probe run. This reduces the need to check that the probe returned a status between 200 - 299.
2. Add the result value to a new prometheus metric: `monika_probe_result`

When using prometheus/grafana, it's easier to have a boolean success or fail metric vs checking that the status code was 200 - 299. It also scales to more probe types, allowing status code to be used for the native interface status/error code.

1.  Created an enum with 3 values: unknown/failed/success (-1/0/1)
2. Each probe adds this result to the `ProbeRequestResponse`
3. The prometheus collector pushes this to the new `monika_probe_result` metric, as well as adds it as a label to `monika_request_response_time_seconds` and `monika_request_response_size_bytes`.

## How to test  
1. `monika -c monika.json --prometheus 3001`
2. View metrics: http://localhost:3001/metrics


src/interfaces/request.ts
```diff
@@ -37,6 +37,13 @@ export type RequestTypes =
   | 'mongo'
   | 'postgres'
 
+// The success/failure result of a probe
+export enum ProbeRequestResult {
```
haricnugraha, line 41: Any problem if it changed to positive integer? Something like 0 (failed), 1 (success), 2 (unknown).

jgillick: No problems here. I've updated the PR

haricnugraha (approved):
LGTM!

sapiderman (approved):
great PR, Thanks!

Issue #1069, opened 2023-06-27
Default maxRedirects too shallow


sapiderman:
**Describe the bug**
1. When probing a url with redirect, monika returns error service down instead of following redirect

**To Reproduce**
1. Use the following config.yaml
```yaml
probes:
  - id: 'http-1'
    name: 'status-200-test'
    requests:
      - url: https://www.spotify.com
    alerts:
      - assertion: response.status < 200 or response.status > 308
        message: HTTP Status is not 200
    incidentThreshold: 3
    recoveryThreshold: 3

notifications:
  - id: desktop
    type: desktop
```
3. Run the probes `monka -c config.yaml`

**Expected behavior**
1. Expect probe to work.
2. Expect some redirects to work

**Screenshots**
![image](https://github.com/hyperjumptech/monika/assets/964106/d45c32fc-4ee2-474a-ab7a-4f78fa7daafb)

**Desktop (please complete the following information):**
 - OS: PopOS 22.0 LTS
 - Monika: `@hyperjumptech/monika/1.15.3 linux-x64 node-v18.16.0`

PR #1070, opened 2023-06-27
Fix default maxRedirect too shallow

sapiderman:
# Monika Pull Request (PR)  

1. Fix issue #1069
2. When probing urls with redirects, monika did not follow and return some error
3. Error MAX redirect from axios was not handled properly
4. Probing with redirect works by default 

1.  Update the default maxRedirect to  3
2. Update docs
3. Handle all possible axios errors

## How to test  
1. Use the same yaml as in the Issue
```yaml
probes:
  - id: 'http-1'
    name: 'status-200-test'
    requests:
      - url: https://www.spotify.com
    alerts:
      - assertion: response.status < 200 or response.status > 2308
        message: HTTP Status is not 200
    incidentThreshold: 3
    recoveryThreshold: 3

notifications:
  - id: desktop
    type: desktop
```
5. Run the probe `npm start -- -c test.yaml`
6. Redirects should work

## before PR
image

## after PR
image


src/context/monika-flags.ts
```diff
@@ -52,7 +52,7 @@ export const monikaFlagsDefaultValue: Pick<
   config: getDefaultConfig(),
   'config-filename': 'monika.yml',
   'config-interval': 900,
-  'follow-redirects': 1,
+  'follow-redirects': 3,
```
haricnugraha, line 55: IMO it is still too shallow. For comparison, curl default is 50. And Axios default is 5. Need to wait others opinion on this.
More info: https://curl.se/docs/manpage.html#--max-redirs and https://axios-http.com/docs/req_config

sapiderman: Sure, what should . Also interesting, axios README.md says their default is 21 [here](https://github.com/axios/axios/blob/21a5ad34c4a5956d81d338059ac0dd34a19ed094/README.md?plain=1#L475)

There may have been attempts to fix timeouts in neosense by limitting to maxRedirect to 1.

sapiderman: setting it to 21 as per axios github docs

haricnugraha (approved):

nicnocquee (approved):

PR #1071, opened 2023-06-27
Fix GitHub Actions Injection Vulnerability

AdnaneKhan:
# Monika Pull Request (PR)  

1. This PR fixes a vulnerability where the issue announcement workflow is vulnerable to injection via a pull request or issue (see https://cycode.com/blog/github-actions-vulnerabilities/). Even though it does not check out code from a PR the workflow has a GITHUB_TOKEN with write access. It is possible to obtain this by dumping the runner's memory. I am happy to prove it if you would like.

1.  The fix is to limit the workflow permissions to read on issues and PRs and to save the titles to environment variables prior to using them in a script run step.

## How to test  
1. To validate the bug, create an issue with the following title `TestIssue" && echo "RCE" && ls -la / && echo "FooBar`

### DIFF
--- a/.github/workflows/teams-notify.yml
+++ b/.github/workflows/teams-notify.yml
@@ -11,23 +11,27 @@ on:
     types: [opened]
   pull_request_target:
     types: [opened, reopened]
-
+permissions:
+  issues: read
+  pull-requests: read
 jobs:
   notify:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


sapiderman (approved):
Thanks, this is a good PR and a reminder to implement the github security recommendtions.

dennypradipta (approved):

PR #1076, opened 2023-07-17
Fix broken Nightly Workflow

sapiderman:
# Monika Pull Request (PR)  

1.  Force PR token to "read" only

1.  Add write permission for PRs

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -7,7 +7,7 @@ on:
 
 permissions:
   issues: read
-  pull-requests: read
+  pull-requests: write
   
 jobs:
   nightly:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 permissions:
   issues: read
-  pull-requests: read
+  pull-requests: write
```
nicnocquee, line 10: i think this one is ok to be `read`. this workflow needs `contents` scope to be `write` i think because there's a step to create and push a new tag to the repository. https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

nicnocquee (changes requested):

.github/workflows/nightly-release.yml
sapiderman: fixed.. added job specific permission for contents.

nicnocquee (approved):

PR #1102, opened 2023-08-18
Fix: Webhook Notification is Not Working

haricnugraha:
# Monika Pull Request (PR)

Fix webhook notification is not working

### Before

https://github.com/hyperjumptech/monika/assets/15191978/322c37c2-7c34-4ac3-bbb9-632bba77c6b6

### After

https://github.com/hyperjumptech/monika/assets/15191978/0673c848-768e-4c79-8f6e-814e94f9c949

## How to test
1. Run a server that print the request body
2. Run Monika with the following configuration

### Webhook notification server example
```json
{
  "name": "webhook-notification",
  "version": "0.0.1",
  "description": "A server to test webhook notification",
  "main": "index.js",
  "scripts": {
    "dev": "nodemon index.ts",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [
    "webhook",
    "monika"
  ],
  "author": "haricnugraha@gmail.com",
  "license": "MIT",
  "dependencies": {
    "express": "4.18.2"
  },
  "devDependencies": {
    "@types/express": "^4.17.17",
    "@types/node": "^20.5.0",
    "nodemon": "^3.0.1",
    "ts-node": "^10.9.1",
    "typescript": "^5.1.6"
  },
  "compilerOptions": {
    "outDir": "./dist"
  }
}
```

```typescript
import type { Request, Response } from "express";
import express from "express";

const app = express();
const port = 3000;

app.use(express.json());

app.post("/", (req: Request, res: Response) => {
  console.log(req.body);

  res.json(req.body);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
});
```

### Monika configuration
```yaml
probes:
  - id: HTTPBin
    requests:
      - url: https://httpbin.org/status/404
    incidentThreshold: 1
    recoveryThreshold: 1
    interval: 5
notifications:
  - type: webhook
    id: webhook-01
    data:
      url: http://localhost:3000
```

### DIFF
--- a/packages/notification/channel/webhook.ts
+++ b/packages/notification/channel/webhook.ts
@@ -36,12 +36,24 @@ export const validator = Joi.object().keys({
 
 export const send = async (
   { url }: NotificationData,
-  { body }: NotificationMessage
+  { meta }: NotificationMessage
 ): Promise<void> => {
+  if (meta.type !== 'incident') {
+    return
+  }
+
+  const { alertQuery, probeID, time } = meta
+
   await sendHttpRequest({
     method: 'POST',
     url,
-    data: body,
+    data: {
+      body: {
+        url: meta.url || probeID,
+        time,
+        alert: alertQuery,
+      },
+    },
   })
 }
 
--- a/packages/notification/channel/webhook.ts
+++ b/packages/notification/channel/webhook.ts
@@ -38,7 +38,7 @@ export const send = async (
   { url }: NotificationData,
   { meta }: NotificationMessage
 ): Promise<void> => {
-  if (meta.type !== 'incident') {
+  if (meta.type !== 'incident' && meta.type !== 'recovery') {
     return
   }
 
--- a/packages/notification/channel/__test__/webhook.test.ts
+++ b/packages/notification/channel/__test__/webhook.test.ts
@@ -1,7 +1,35 @@
+/**********************************************************************************
+ * MIT License                                                                    *
+ *                                                                                *
+ * Copyright (c) 2021 Hyperjump Technology                                        *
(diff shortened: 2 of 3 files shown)


nicnocquee (approved):

packages/notification/channel/webhook.ts
```diff
@@ -36,12 +36,24 @@ export const validator = Joi.object().keys({
 
 export const send = async (
   { url }: NotificationData,
-  { body }: NotificationMessage
+  { meta }: NotificationMessage
 ): Promise<void> => {
+  if (meta.type !== 'incident' && meta.type !== 'recovery') {
```
raosan, line 41: why add this condition? how about the start messsage with `meta.type === 'start'`?

haricnugraha: I assume that the webhook notification is created when there is only incident type. Because, if you read [the documentation](https://monika.hyperjump.tech/guides/notifications#webhook), it only sends the following data:

```json
{
 "body": {
   "url": "string"
   "time": "string"
   "alert": "string"
  }
}
```

Only notification with incident or recovery types that have url, time, and alert data. So, I added the condition. Otherwise, it will send empty data.

Of course, we need to improve the webhook data in the future to send other type of notification. But this PR only fixes the existing feature.

raosan (approved):

packages/notification/channel/webhook.ts
raosan: i think one of the most basic and important is the `start notification` to know that the notification is working. It should be next in line

PR #1116, opened 2023-09-05
fix: monika install

raosan:
# Monika Pull Request (PR)

## What feature/issue does this PR add
1. cannot install monika via npm

## How did you implement / how did you fix it
1. change package.json so that monika-notificaiton is installed from npm

## How to test
1. npm i
2. npm run test
3. npm run start

PR #1117, opened 2023-09-05
Update for release v1.15.8

raosan:
# Monika Pull Request (PR)

1. Update for release v1.15.8

### Fixed

- Monika Install (#1116)


package.json
sapiderman, line 1: line 63 need to be updated?

raosan: no need, since version 1.15.7 is the one on npm. If we update, then it will throw error

PR #1139, opened 2023-10-13
Fix: Increase Request Timeout When Getting Probes from Symon

haricnugraha:
# Monika Pull Request (PR)

This PR increases the request timeout when getting probes from Symon, which caused Monika to stop probing due to not receiving any configuration.

## How did you implement / how did you fix it?
- Increased the request timeout from 10 seconds to 30 seconds.
- Let it to throw an error if the request fails to ensure that it is noticeable.

## How to test?
Assign a large number of probes to Monika, that will causing the request to fetch probes to take longer than 10 seconds.

### Before

https://github.com/hyperjumptech/monika/assets/15191978/bf394740-f1fa-48a0-a8c4-3f9702ff78d5

### After

https://github.com/hyperjumptech/monika/assets/15191978/a80ea47c-de57-4c13-8ad6-d40a07b0472d


src/symon/index.ts
```diff
@@ -325,8 +325,11 @@ class SymonClient {
 
   private async fetchProbes() {
     log.debug('Getting probes from symon')
+    const TIMEOUT = 30_000
+
     return this.httpClient
       .get<{ data: Probe[] }>(`/${this.monikaId}/probes`, {
+        timeout: TIMEOUT,
```
dennypradipta, line 332: Just a simple request, could you make it override-able? So that we can test it in the Symon using a flag to know which sweet spot for each location

dennypradipta (commented):
A simple improvements, might wanna check

src/symon/index.ts
haricnugraha: This PR is a workaround. I'm currently working on the long term solution.

sapiderman (approved):
LGTM

ilmiawan (approved):

PR #1159, opened 2023-11-05
Fix: Add handled stopsignal to Dockerfile

saydamir:
# Monika Pull Request (PR)

1. Currently, Monika only handles the SIGINT signal for graceful shutdown: https://github.com/hyperjumptech/monika/blob/c0ec3df5888848c385e5922eb51abf021d01e4b5/src/commands/monika.ts#L454
Docker uses the `SIGTERM` signal by default. If the application does not stop within the given timeout, the Docker will terminate it by sending the `SIGKILL` signal.

1. Add `STOPSIGNAL SIGINT` to Dockerfile.

## How to test
1. Build docker image:  
`docker build . -t monika`
3. Run it:  
`docker run -d --rm --name monika monika`
4. Stop:  
```
time docker stop --time 15 monika  
monika
docker stop --time 15 monika  0.04s user 0.03s system 0% cpu 15.741 total
```
5. The stop command will take 15 seconds if a signal is sent that is not handled by the application.


dennypradipta (approved):
LGTM. I tested it and it works fine. I tried to run a container without `-d` flag, and pressed CTRL+C and it still works. Thank you for your PR!

sapiderman (approved):
Looks good. Thanks for the PR.

Issue #1164, opened 2023-11-07
Repeat 1 doesn't exit when the probe is not accessible


nicnocquee:
**Describe the bug**
Monika doesn't exit when the probe in the configuration is not accessible. Instead, it keeps retrying.

**To Reproduce**
1. Run monika from the main branch: `npm run start -- -c ./monika.yml -r 1`
2. Use the following configuration:

```yaml
probes:
  - id: '1'
    name: Localhost
    description: Multiple
    requests:
      - url: http://localhost:3000
```

**Expected behavior**
Monika should exit even when the probe cannot be reached. 

**Screenshots**
![image](https://github.com/hyperjumptech/monika/assets/311343/6102bd38-5485-4c18-9bb8-be1513123355)

**Desktop (please complete the following information):**
 - OS: macOS Sonoma 14.0 (23A344)
 - Version 72302026976793f23f58e10aabff6ac4baf57b07
 - Node v18.17.1

dennypradipta:
Should be closed by #1165

PR #1165, opened 2023-11-07
Fix repeat flag doesn't terminate monika on fail probe

nicnocquee:
# Monika Pull Request (PR)

1. #1164 

There are several bugs that causes the issue #1164 

1. The `setProbeRunning` is called almost every second in this line https://github.com/hyperjumptech/monika/pull/1165/files#diff-c3cb2be3e6fbad9e8455e2fc282a2d52cc7875688162e517023bff5be26371b6L54. The fix: moved it back down as originally written.
2. The `retry` function is called without `maxAttempt` here https://github.com/hyperjumptech/monika/pull/1165/files#diff-c3cb2be3e6fbad9e8455e2fc282a2d52cc7875688162e517023bff5be26371b6R81. This causes Monika to keep retrying to monitor the probe in a cycle which causes the bug in #1164. The fix: added `maxAttempt` with value whichever bigger between `incidentThreshold` and `recoveryThreshold` because we want to retry for both incident and recovery cases. See below for more information.
3. Removal of `recoveryThreshold` causes the recovery notification to be sent immediately upon recovery. But we don't want that. Just like we don't want to immediately send incident notification, Monika needs to retry several times first before sending the recovery notification. The fix: 

    1. restore the `recoveryThreshold` and remove the deprecation warning https://github.com/hyperjumptech/monika/pull/1165/files#diff-028bd48d9206d28afc908d305ef5ac57026ad51ba014a0741fb8abcaef2af741R113. I think there was misunderstanding, we just don't want user to be able to set the threshold from symon, but we still need it in Monika. 
    2. don't immediately send the recovery notification, instead keep retrying until `recoveryThreshold`. https://github.com/hyperjumptech/monika/pull/1165/files#diff-537950d0505c20b7b3ec67590102fdafdbbbbddb0130d5dbf7a4ac232060598dR152

Some further changes to improve performance:

1. Set the default `retryInitialDelayMs` value to 2 seconds. 128 milliseconds is not a sensible default. There's no need for Monika to retry that fast. https://github.com/hyperjumptech/monika/pull/1165/files#diff-06df4a808fa037b48c082497d352cb7825c883648d52a17ab72b0e7dd70a5c5dR84
2. Remove unnecessary error throws, e.g., https://github.com/hyperjumptech/monika/pull/1165/files#diff-5d5a93a3e7827f41dca4b896d385bc9ab744774dd26149e2f605fbe064398021L134

### Probing Mechanism

With this PR, I also fixed the probing mechanism. This is how it works.

Say we have the following probe:

```yaml
probes:
  - id: '1'
    name: Localhost
    description: Check status
    interval: 5
    incidentThreshold: 3
    recoveryThreshold: 3
    requests:
      - method: GET
        timeout: 10000
        url: http://localhost:8000/api/delay
    alerts:
      - assertion: response.time > 800
        message: Response time > 800
```

Then say the server in the localhost:8000 is offline (unreachable). The process will go like the following:

1. In the first cycle, Monika will try to probe it but fails. Monika will not send the incident notification immediately. Instead, it will retry 3 times (based on the incidentThreshold) with exponential backoff algorithm. 
2. If in the second retry, the server is up, then it will do nothing and just wait for 5 seconds (based on the interval) to start another cycle. 
3. On the other hand, if after 3 attempts the server is still unreachable, it will send the incident notification, and mark the probe as has incident. Then it will wait until 5 seconds (interval) has passed to probe it again.
4. In the next cycle, if the server is online, Monika will not send the recovery notification immediately because the probe was marked as incident previously. Instead, it will retry to monitor the probe 3 times (according to recoveryThreshold) before sending the recovery notification. 

The same process is also executed when the probe can reached but the assertion fails which in this case when response's time is greater than 800 ms.

When the `-r` flag is provided, Monika will perform as many cycles as defined by the repeat flag value. However, it will also still do retries in case of incident and recovery as defined by the probes' `incidentThreshold` and `recoveryThreshold` if provided, or use their respective default values.

## How to test
1. Use the following monika.yml:

```yaml
probes:
  - id: '1'
    name: Localhost
    description: Multiple
    requests:
      - url: http://localhost:3000
```

6. `npm run start -- -c ./monika.yml -r 2`

## Screenshots

Running with `-r 1` and the probe cannot be reached (offline):

Screenshot 2023-11-08 at 13 24 28

Running with `-r 2` and the probe cannot be reached (offline):

Screenshot 2023-11-08 at 13 24 40

Running with `-r 1` and the probe can be reached (online):

Screenshot 2023-11-08 at 13 27 32

Running without `-r` flag (continuously):

Screenshot 2023-11-08 at 13 27 11


dennypradipta:
I can't seem to reproduce the first case where the config is monika.yml with `-r 1` flag. I still have the incident threshold as 5 instead of 2:
image

But after the probe is marked as incident Monika stopped itself though. Was it intended or not?

nicnocquee:
The first and second cases are by design. The fourth case should be fixed now. 

image

dennypradipta (approved):
LGTM

sapiderman (approved):
Worked for me. Ready to merge.

PR #1185, opened 2023-11-16
Feat: Add --ignoreInvalidTLS flag

### LINKED ISSUE #1182: Allow users to set rejectUnauthorized on HTTPS Agent from CLI flags
In httpRequest function, it creates an https agent that does not have the ability to set the rejectUnauthorized value. It is configurable through the request config, but it would be great if it can be set through the CLI flags

dennypradipta:
# Monika Pull Request (PR)

This PR fixes #1182 

1. Add new flag: `--ignoreInvalidTLS`
2. If the flag is present, it will set `rejectUnauthorized` to the inverse value of `ignoreInvalidTLS` flag.
3. `allowUnauthorized` property in a probe still takes higher priority than the global flag

## How to test

Prepare the config:
```yml
notifications:
  - id: 'desktop'
    type: desktop
probes:
  - id: '1'
    name: 'Coding Horror'
    description: 'Self signed badssl'
    interval: 3 # in seconds
    requests:
      - method: GET
        url: 'https://self-signed.badssl.com'
        # allowUnauthorized: true
    alerts: []
``

1. Run `npm run start -- -c config.yml --ignoreInvalidTLS`. It should not display error.
2. Run `npm run start -- -c config.yml`. It should display error.
3. Uncomment the `allowUnauthorized` in the config file and run `npm run start -- -c config.yml`. It should not display error.

## Video


https://github.com/hyperjumptech/monika/assets/16670475/4ac37706-a523-4ac5-bbc1-f3afd3c1faea

### DIFF
--- a/docs/src/pages/guides/cli-options.md
+++ b/docs/src/pages/guides/cli-options.md
@@ -320,6 +320,14 @@ monika --summary
 Please note that you need to run the above command from the same working directory as the running monika you want to see/check.
 The '--summary' flag also will not work when no other monika process is running.
 
+## TLS Reject Unauthorized
+
+If there is a probe with request(s) that uses HTTPS, Monika will show an error if the target's TLS certificate is invalid. You can configure whether HTTPS requests should ignore invalid certificates using the `--ignoreInvalidTLS` flag.
+
+```bash
+monika --ignoreInvalidTLS
+```
+
 ## Verbose
 
 Like your app to be more chatty and honest revealing all its internal details? Use the `--verbose` flag.
--- a/src/commands/monika.ts
+++ b/src/commands/monika.ts
@@ -135,6 +135,12 @@ export default class Monika extends Command {
       multiple: false,
     }),
 
+    ignoreInvalidTLS: Flags.boolean({
+      description:
+        'Configures whether HTTPS requests should ignore invalid certificates',
+      default: false,
+    }),
+
     insomnia: Flags.string({
       char: 'I', // (I)nsomnia file to
       description: 'Run Monika using an Insomnia json/yaml file',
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: docs/src/pages/guides/cli-options.md [hunk 1]
```diff
@@ -320,6 +320,14 @@ monika --summary
 Please note that you need to run the above command from the same working directory as the running monika you want to see/check.
 The '--summary' flag also will not work when no other monika process is running.
 
+## TLS Reject Unauthorized
+
+If there is a probe with request(s) that uses HTTPS, Monika will show an error if the target's TLS certificate is invalid. You can configure whether HTTPS requests should ignore invalid certificates using the `--ignoreInvalidTLS` flag.
+
```
sapiderman, line 326: do we know what kind of "invalid" certificates are occurring? can we be more specific if possible? All invalid certs will be ignored?

dennypradipta: Accordin to the [NodeJS docs](https://nodejs.org/api/tls.html#tlsconnectoptions-callback):

>rejectUnauthorized [<boolean>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type) If not false, the server certificate is verified against the list of supplied CAs. An 'error' event is emitted if verification fails; err.code contains the OpenSSL error code. Default: true.

Some of the problems occurred is, but not limited to:
- Error: certificate has expired
- DEPTH_ZERO_SELF_SIGNED_CERT Error: self-signed certificate
- ERR_TLS_CERT_ALTNAME_INVALID Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames: Host: somehost.com. is not in the cert's altnames: DNS:*.dns.any, DNS:dns.any
- Error: write EPROTO 589B604DD17F0000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:354:

And many more that I didn't know much about it.

dennypradipta: I do found some research though from ChatGPT but it seems that some of them differ from the Node.js ones

https://github.com/openssl/openssl/blob/master/crypto/err/openssl.txt

dennypradipta: Found the docs. Updated the commit

sapiderman (approved):
good PR! 👍🏽 👍🏽 👍🏽  ready to go

ilmiawan (approved):

PR #1188, opened 2023-11-20
Release: Upgrade the Application Version to 1.16.0 and Update the Changelog

haricnugraha:
# Monika Pull Request (PR)

- Upgrade the application version to 1.16.0
- Update the Changelog

### DIFF
--- a/packages/notification/package.json
+++ b/packages/notification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperjumptech/monika-notification",
-  "version": "1.15.11",
+  "version": "1.15.12",
   "description": "notification package for monika",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
--- a/packages/notification/package.json
+++ b/packages/notification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperjumptech/monika-notification",
-  "version": "1.15.11",
+  "version": "1.16.0",
   "description": "notification package for monika",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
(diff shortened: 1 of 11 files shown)


sapiderman (approved):
good to go!

packages/notification/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hyperjumptech/monika-notification",
-  "version": "1.15.11",
+  "version": "1.16.0",
```
raosan, line 3: actually no need to update it anymore on every monika release.
according to [this](https://github.com/hyperjumptech/monika/tree/main/packages/notification#releasing), monika/notification should be updated individually every PR related to `packages/notification`
```suggestion
  "version": "1.15.11",
```

raosan: instead, we need to update the version in `package.json` file line 66
```
"@hyperjumptech/monika-notification": "1.15.7", // update to latest version 1.15.11
```

haricnugraha: Thank you for the information. The latest version of `@hyperjumptech/monika-notification` is `1.15.7`, but there are some code changes after that release. I think it would be an operational burden to separate the releases. IMO, it would be easier if we just do one release and use one version. But, we should change the release process to release the package in the workspace first, then release the root package. 

## NPM
https://www.npmjs.com/package/@hyperjumptech/monika-notification?activeTab=versions

image

## GitHub
https://github.com/hyperjumptech/monika/commits/main/packages/notification

image

raosan: just need to update monika-notification version in root package.json, then resolve this comment

raosan (approved):

dennypradipta (approved):
Please edit the CHANGELOG to inform about #1190

packages/notification/package.json
haricnugraha: Ok, updated. Thank you for the discussion.

PR #1203, opened 2023-12-04
Refactor: Update Probe Data Independently

### LINKED ISSUE #1163: Optimize Probe Update Mechanism
I propose a better mechanism for Monika to receive probe data from Symon. There are significant changes on both side. Here are the steps:

1. Every time a probe changes (`add/update/delete`), a record will be stored in the `probe_changes` table.
2. Monika will continue to poll to Symon to get probe data. But, instead of retrieving all the probes, it will only get probe change data from a specific time.
3. When there are probe changes, Monika will take the following actions:

- `add` changes: Add the probe to the scheduler.
- `remove` changes: Remove the probe from the scheduler.
- `update` changes: Remove the probe from the scheduler and then re-add it.

The requirement for this mechanism to work is that Monika needs to change its mechanism to use a scheduler that is independent for each probe.

## Diagram
probe-getter

haricnugraha:
# Monika Pull Request (PR)

Update probe data independently. It is an effort to resolves #1163.

1. Polling to Symon probe changes endpoint to get probe changes data.
2. Apply the probe changes based on the changes type.

## How to test

1. Run Monika in Symon mode.
2. Add, update, delete a probe.
 
### Video

https://github.com/hyperjumptech/monika/assets/15191978/766a992f-f193-4926-84a0-003aa1bb6772

### DIFF
--- a/src/components/config/index.ts
+++ b/src/components/config/index.ts
@@ -98,7 +98,7 @@ export const updateConfig = async (config: Config): Promise<void> => {
 
     setContext({ config: newConfig })
     setProbes(newConfig.probes)
-    emitter.emit(events.config.updated, newConfig)
+    emitter.emit(events.config.updated)
     log.info('Config file update detected')
   } catch (error: any) {
     if (isTestEnvironment) {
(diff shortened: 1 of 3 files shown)


sapiderman (approved):
👍🏽 looks okl RTM.

lukman7788 (approved):

PR #1214, opened 2023-12-15
Feat: Probing with Queue

### LINKED ISSUE #980: Investigate probing mechanism with Bree or similar
Since we already use Bree for reporting to Symon, maybe we can also use probing probes using Bree.

dennypradipta:
# Monika Pull Request (PR)

This PR fixes #980 

Even though we did not use Bree or Piscina as it is too much effort to refactor, I tried to do a POC using https://www.npmjs.com/package/queue to queue the probing functions. 

Supposedly there are 800 probes, when it is time to probe, it will be queued.

## How to test

Run Monika with Symon Mode


haricnugraha:
Could you elaborate more on the differences before and after implementing the queue? I think you could achieve the same result by removing the setTimeout function.

lukman7788 (approved):

ilmiawan (approved):

Issue #1216, opened 2023-12-19
Error "Invalid URL" after deleting monika-logs.db


kevinhermawan:
After deleting `monika-logs.db`, it shows an error `ERROR: Warning: Can't open logfile. Invalid URL`.

![The error](https://github.com/hyperjumptech/monika/assets/84965338/33523023-b9d0-4180-8b75-798003c0818a)

PR #1218, opened 2023-12-19
Refactor: Update Probe in Symon Mode Without Restarting

### LINKED ISSUE #1163: Optimize Probe Update Mechanism

haricnugraha:
# Monika Pull Request (PR)

Update probe in Symon mode without restarting. It will resolve #1163.

## How did you implement / how did you fix it?

1. Always get the latest probe data from the cache for each iteration.
2. Update the local probe state when there are probe data changes.

## How to test:

1. Run Monika in Symon mode.
2. Add, update, and delete a probe.
3. Run Monika in Monika mode.

#### Symon Mode

https://github.com/hyperjumptech/monika/assets/15191978/a50e16cf-cd04-4905-9ec0-30df57c040bd

#### Monika Mode (Regression Test)

https://github.com/hyperjumptech/monika/assets/15191978/ab0e5a39-acbf-4c0e-b168-605b36c338a9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
         "packages/*"
       ],
       "dependencies": {
-        "@breejs/ts-worker": "^2.0.0",
         "@faker-js/faker": "^7.4.0",
         "@hyperjumptech/monika-notification": "^1.15.13",
         "@oclif/core": "2.15.0",
@@ -22,7 +21,6 @@
         "ajv": "^8.11.0",
         "axios": "^0.27.2",
         "boxen": "^5.0.0",
-        "bree": "^9.1.3",
         "chalk": "^4.1.2",
         "chokidar": "^3.5.1",
         "cli-table3": "^0.6.0",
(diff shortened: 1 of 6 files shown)


kevinhermawan (approved):

haricnugraha:
> Thanks, it also fixes #1216

I have separated the PR for this issue in #1219. Previously, I included the fix here to make it possible to test.

sapiderman (approved):
good to go!

PR #1233, opened 2024-01-19
Fix: Monika Multi Node

haricnugraha:
# Monika Pull Request (PR)

Implement a multi-node Monika with a probe changes mechanism to update the probe.

The implementation checks for probe assignment data changes. If there are changes, Monika will re-fetch probe configuration data.

## How to test

Run multiple instances of Monika in Symon mode. Run `npm start -- --symonUrl=[symonUrl] --symonKey=[symonKey] --symonLocationId=[symonLocationId] --symon-api-version=v2 --symonGetProbesIntervalMs=5000`.

## Video

https://github.com/hyperjumptech/monika/assets/15191978/0ef55d5a-eb75-4472-9294-d2aeb5919f56

### DIFF
--- a/src/symon/index.ts
+++ b/src/symon/index.ts
@@ -153,7 +153,6 @@ export default class SymonClient {
   private worker
   private apiKey: string
   private probeChangesInterval: NodeJS.Timeout | undefined
-  private hasConnectionToSymon: boolean = false
   private httpClient: AxiosInstance
   private locationId: string
   private monikaId: string
@@ -238,7 +237,6 @@ export default class SymonClient {
     }, getContext().flags.symonGetProbesIntervalMs)
 
     this.report().catch((error) => {
-      this.hasConnectionToSymon = false
       log.error(`[Symon] Report failed. ${(error as Error).message}`)
     })
 
@@ -249,7 +247,7 @@ export default class SymonClient {
   }
 
   async sendStatus({ isOnline }: { isOnline: boolean }): Promise<void> {
-    const { status } = await this.httpClient({
+    await this.httpClient({
       data: {
         monikaId: this.monikaId,
         status: isOnline,
@@ -257,10 +255,6 @@ export default class SymonClient {
       method: 'POST',
       url: '/status',
     })
-
-    if (status === 200) {
-      this.hasConnectionToSymon = true
-    }
   }
 
   async stop(): Promise<void> {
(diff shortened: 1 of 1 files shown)


dennypradipta:
I tried testing this PR, is it normal to have 3 locations have 3 exact probes?

image

As you can see from the terminal above, I have run each terminal with this command:
`npm start -- --symonUrl=http://localhost:3000 --symonKey=<key> --symonLocationId=<loc_id> --symon-api-version=v2 --symonGetProbesIntervalMs=5000`

haricnugraha:
> I tried testing this PR, is it normal to have 3 locations have 3 exact probes?
> 
> image
> As you can see from the terminal above, I have run each terminal with this command: `npm start -- --symonUrl=http://localhost:3000 --symonKey=<key> --symonLocationId=<loc_id> --symon-api-version=v2 --symonGetProbesIntervalMs=5000`
> 
> I got https://netflixtechblog.com in each instance. Symon is using feat/probe-assignment-last-updated-at branch

It looks like you are running three Monika on the same node. Symon will assume it's one Monika based on the MAC address. You can check this in the operator dashboard. This will work if you have a different node with a different MAC address or if you comment out the Symon code that separates the Monika based on their MAC addresses.

sapiderman (approved):

src/symon/index.ts
```diff
@@ -286,6 +259,40 @@ export default class SymonClient {
     this.probeChangesCheckedAt = probeChangesCheckedAt
   }
 
+  private async fetchAndApplyProbeChanges() {
```
raosan, line 262: it works on my machine 🦾 🤖 

btw i think i found an edge case. I don't know if this is covered in this task.

How to reproduce:
1. run symon, create monitor with location A.
2. run monika using your described config with loaction B
3. change the location of created monitor in symon to location B
4. the config is not detecting change

raosan (approved):

dennypradipta (approved):
Nicely done. Working on my machine
Screenshot 2024-01-30 at 10 30 44

But it seems there are some monika instances that does not have any probe unless updated. I guess there is something in the Symon API rather than Monika codebase

src/symon/index.ts
haricnugraha: It needs to be fixed in the Symon. I will try to fix it later.

PR #1245, opened 2024-03-06
Release: Version 1.18.0

haricnugraha:
# Monika Pull Request (PR)

Release version 1.18.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@hyperjumptech/monika",
   "description": "Synthetic monitoring made easy",
-  "version": "1.17.0",
+  "version": "1.18.0",
   "license": "MIT",
   "author": "@hyperjumptech",
   "main": "lib/index.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@hyperjumptech/monika",
   "description": "Synthetic monitoring made easy",
-  "version": "1.17.0",
+  "version": "1.18.0",
```
sapiderman, line 4: just a toggle is not just an update to patch number (ex:1.17.1)? 😁

sapiderman (approved):

kevinhermawan (approved):

package.json
haricnugraha: According to the [Semantic versioning document](https://semver.org/), the minor version needs to be incremented if there is a new backward-compatible functionality. My thinking is, it is still a new functionality even though it is doing nothing.

PR #1278, opened 2024-04-19
[Feat] Removing Axios for Native Fetch

sapiderman:
# Monika Pull Request (PR)

1. Implements #1248 
2. Replacing with native fetch main Monika http transactions

1. Remove sendHttp wrappers with native fetch

## How to test

1. `npm start -- -c test.yaml` should work
2. `npm test` should all pass
4. `monika -c https://raw.githubusercontent.com/hyperjumptech/monika/main/config_sample/config.desktop.example.yml` should be fetched properly
5. `monika --auto-update patch` should work


src/plugins/updater/index.ts
```diff
@@ -268,14 +270,19 @@ async function downloadMonika(
...
   const downloadUri = `https://github.com/hyperjumptech/monika/releases/download/v${remoteVersion}/${filename}.zip`
   log.info(`Updater: download from ${downloadUri}`)
-  const { data: downloadStream } = await sendHttpRequest({
+  const resp = await sendHttpRequestFetch({
     url: downloadUri,
     responseType: 'stream',
   })
 
-  const { data: checksum }: { data: string } = await sendHttpRequest({
+  const { data: downloadStream } = (await resp.json()) as { data: Stream }
```
syamsudotdev, line 278: I think we can get stream from `resp.body` instead of `resp.json()` ?

ref https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams

src/symon/index.ts
```diff
@@ -296,49 +296,59 @@ export default class SymonClient {
...
-        )
-      })
-      .catch((error) => {
-        log.error(
-          `[Symon] Send ${
-            probeState === 'DOWN' ? 'incident' : 'recovery'
-          } event for Alert ID: ${alertId} failed.  ${(error as Error).message}`
-        )
+    try {
+      const resp = await sendHttpRequestFetch({
```
syamsudotdev, line 307: I guess this API call is missing `x-api-key` header ?

maybe we can create a reusable function inside symon client, something like
```
private fetcher() {
  sendHttpRequestFetch({
    // base url
    // common headers
  })
}
```

src/symon/index.ts
```diff
@@ -361,61 +371,49 @@ export default class SymonClient {
...
-        if (!res.data.data) {
-          return {
-            probes: getProbes(),
-            hash: res.headers.etag,
-          }
-        }
+    const resp = await sendHttpRequestFetch({
+      url: `${this.url}/api/${this.apiVersion}/${this.monikaId}/probes`,
+      method: 'GET',
+      headers: {
```
syamsudotdev, line 394: another one with no x-api-key header

src/symon/index.ts
```diff
@@ -440,9 +438,17 @@ export default class SymonClient {
...
     }
 
-    return this.httpClient
-      .post('/client-handshake', handshakeData)
-      .then((res) => res.data?.data.monikaId)
+    const resp = await sendHttpRequestFetch({
+      url: `${this.url}/api/${this.apiVersion}/monika/client-handshake`,
+      method: 'POST',
+      body: JSON.stringify(handshakeData),
+      headers: {
```
syamsudotdev, line 445: another one without `x-api-key` header

src/components/config/parser/insomnia.ts
```diff
@@ -110,10 +109,14 @@ export function parseInsomnia(configString: string, format: string): Config {
     ? `${env?.scheme?.[0] ?? 'http'}://${env?.host}${env?.base_path}`
     : ''
 
-  return mapInsomniaToConfig(data)
+  return mapInsomniaToConfig(data, environmentVariables, baseUrl)
 }
 
-function mapInsomniaToConfig(data: unknown): Config {
+function mapInsomniaToConfig(
```
ilmiawan, line 115: for effectivity mapInsomniaToConfig and mapInsomniaRequestToConfig can be merged. By making mapInsomniaToConfig as a pure function that return insomniaRequests array, then call it from inside mapInsomniaRequestToConfig rather than just passing parameters to the next function.

syamsudotdev:
@sapiderman theres snapcraft-credential file committed

PR #1324, opened 2024-11-29
Fix: Fallback for getOSName

dennypradipta:
# Monika Pull Request (PR)

1. Add fallback for getOSName so that it does not crash if fails to fetch the os name

1. Wrapped the function using try-catch
2. Add fallback OS Name

## How to test

1. Run npm run start with notifications enabled. You should see the OS Name
3. Build Docker (`docker build . -t monika`) and run it `docker run -v ./monika.yml:/config/monika.yml --name monika monika monika --config /config/monika.yml`. Be advised that notifications MUST be enabled. You should see the OS Name

## Images
image
image

## Additional Context

My machine is detected as macOS unknown, this is caused by old version of `os-name`. While I try to upgrade it, it didn't work because the newest version is ESM only, and you know how it goes with ESM modules for Monika... 😅

### DIFF
--- a/src/components/notification/alert-message.ts
+++ b/src/components/notification/alert-message.ts
@@ -221,14 +221,21 @@ export async function getOSName(): Promise<string> {
   const isLinux = osPlatform === 'linux'
 
   if (isLinux) {
-    const linuxDistro = await getLinuxDistro()
+    try {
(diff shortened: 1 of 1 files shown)


haricnugraha (approved):

sapiderman (approved):
thank you

